PR #64, opened 2019-05-02
Fix booth function spec in README.md

Balandat:
Not sure what this was, but it wasn't a function...


tejaslodaya:
Yes @Balandat ,
That is confusing. Please merge it to master.

PR #72, opened 2019-05-06
Fixed one_hot.py

carefree0910:
This is a hot fix for `is_valid_type` checking in ./core/search_space.py:
```python
if not self._parameters[name].is_valid_type(value):
    if raise_error:
        raise ValueError(
            f"{value} is not a valid value for "
            f"parameter {self._parameters[name]}"
        )
    return False
```
Since np.str_ couldn't pass the type checking


sdsingh:
Thanks for the PR! This looks good to me, will merge.

PR #102, opened 2019-05-29
Remove unused input

nmasahiro:
Hi! I removed the argument `dtype` from train function because dtype is not used.


ax/utils/tutorials/cnn_utils.py
```diff
@@ -93,10 +93,7 @@ def load_mnist(
 
 
 def train(
-    train_loader: DataLoader,
-    parameters: Dict[str, float],
-    dtype: torch.dtype,
```
c-bata: I guess this argument should be passed to `inputs.to()` at L128.
https://github.com/facebook/Ax/blob/ca3e654224642534c434ee111224aaa93b3a34e1/ax/utils/tutorials/cnn_utils.py#L128

nmasahiro: Thanks! I close this pull request and create new pull request that addresses it.

PR #103, opened 2019-05-29
add dtype argument to inputs

nmasahiro:
I create this pull request instead of #102.


ax/utils/tutorials/cnn_utils.py
```diff
@@ -125,7 +125,7 @@ def train(
     # Train Network
     for inputs, labels in train_loader:
         # move data to proper dtype and device
-        inputs = inputs.to(device=device)
+        inputs = inputs.to(dtype=dtype, ddevice=device)
```
c-bata: As the same with L162, It seems `device` is correct.

PR #110, opened 2019-06-04
[WIP] Add deploying website to CI

JoelMarcey:
Using Docusaurus and the Docusaurus Bot

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,20 +52,11 @@ matrix:
         # since -W flag breaks on first error, re-run without it to print all
         - sphinx-build sphinx/source sphinx/build
 
-deploy_website:
-  language: node_js
-  node_js:
-    - '8'
-  branches:
-    only:
-      - master
-  cache:
-    yarn: true
-  script:
-    - git config --global user.name "${GH_NAME}"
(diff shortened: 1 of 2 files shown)


JoelMarcey:
This is currently a WIP - we need to add a call to the script that does the API docs, versioning, etc. I believe.

kkashin:
@JoelMarcey - this won't run at the moment on Travis since deploy_website is not a supported directive, right? Just want to make sure we don't wipe what's on gh-pages

JoelMarcey:
@kkashin - no. `gh-pages` has not been updated as a result of this PR.

kkashin:
Cool, let's import and get this in then.

JoelMarcey:
@kkashin So, if this actually lands as a commit to `.travis.yml`, then it *will* run as stated and call `yarn run gh-pages`. Do you want me to comment that part out for now? That's why I didn't want to land this yet until we got everything in order with the call to your script(s). But if I comment out that line, we can land it and then iterate via another PR. Either way.

kkashin:
Ah, ok - let's either comment it out or just iterate on this PR until the scripts are called - either one!

PR #117, opened 2019-06-20
Update botorch.py

dgiannelli:
Just fixed a typo in the dictionary get call call that builds the option list which has to be passed to the acquisition function.
That typo has the effect of always passing no options to the acquisition function.


lena-kashtelyan:
Thank you so much, @dgianelli! Will merge your PR shortly.

Issue #133, opened 2019-08-01
how to optimize on integer-valued parameters?


overshiki:
Hi,
I'm just brand new to this field.
Following the tutorial, I could very easily do the optimization on float-valued parameters such as learning rate. However, I couldn't find any guidance on integer-valued parameters, such as the number of layers and the number of neurons per layer.
A brief search lead me to the paper: https://arxiv.org/pdf/1706.03673.pdf which described 3 strategies:
    1. optimize the float valued acquisition function and then wrap the result into the closest integer, before the evaluating step
    2. optimize the float valued acquisition function, use this value as input to the evaluating function, then do the wrapping inside the evaluating function
    2. do the wrapping of the input when calculating the covariance function

I am not sure if Ax or botorch has implemented any interface for integer inputs. If so, which strategy is used? or any other ideas are recommended here?

Thanks a lot in advance.

sdaulton:
Hi @overshiki, you can specify that a `RangeParameter` is an integer using [`ParameterType.INT`.](https://ax.dev/versions/latest/api/core.html#ax.core.parameter.ParameterType). E.g. `RangeParameter(name="x", parameter_type=ParameterType.INT, lower=1, upper=10)`. By default, Ax will apply an [`IntToFloat`](https://ax.dev/versions/latest/api/modelbridge.html#ax.modelbridge.transforms.int_to_float.IntToFloat) transform on generated candidates that basically does (1). You can see all the default transforms [here](https://ax.dev/versions/latest/docs/models.html#transforms). If you are generating batches of new candidates and you are using sequentially (greedy) selecting each point in the batch (the default in Ax), where each candidate in the batch is sequentially selected and subsequent candidates are conditioned upon the already selected points, then each new candidate is rounded before the successive conditioning to improve the optimization performance.

danielrjiang:
Also, you can see how the transformation / untransformation happens before and after a point is generated here, similar to how you describe in (1): https://github.com/facebook/Ax/blob/951aa7bc8ccea2ddc6c0dc18fa8190effef046c3/ax/modelbridge/base.py#L491-L522

PR #134, opened 2019-08-01
clarify how integer RangeParameters are rounded

sdaulton:
Summary: see title

Differential Revision: D16604103


sdaulton:
Addresses #133

PR #144, opened 2019-08-10
Update typeutils.py

carefree0910:
dtypes `np.float32`, `np.int32`, ... is not under consideration, fixed it by calling general `isinstance`
+ Without fixing it, Ax on Windows [Anaconda Distribution] will raise tons of bugs due to `json.dump`


Balandat:
Great, thanks for catching this. Will merge in shortly.

PR #216, opened 2019-11-26
Make `overwrite_existing_experiment` in `create_experiment` fail if experiment set locally

### LINKED ISSUE #215: [Feature]  Make `overwrite_existing_experiment`  in `create_experiment` fail if experiment set locally
Given that AxClient has a 1:1 relationship with experiments, I would really like to see the `create_experiment` function to fail if the local instance has an experiment set. This would be really useful for error runs of JupyterLab notebooks where the `create_experiment` is in a single cell and you may trigger running it again by error. One way to do it would be to extend the `overwrite_existing_experiment` flag to also make the function fail if the local instance has an experiment. We could check the local experiment name being the same as the created one. An extra check would be if the _name is not set for either one of them, and make the function fail as well.

casassg:
Overwrite flag on `AxClient.create_experiment` should check for local experiments in addition  of DB experiments. This helps prevent accidental experiment deletion for use cases where the internal DB is not used as the persistent layer.

I added a test as I think it should behave like. Not sure if passes or not as I did not have it set up to run tests locally.

Closes #215

### DIFF
--- a/ax/service/ax_client.py
+++ b/ax/service/ax_client.py
@@ -188,10 +188,9 @@ class AxClient:
             status_quo: Parameterization of the current state of the system.
                 If set, this will be added to each trial to be evaluated alongside
                 test configurations.
-            overwrite_existing_experiment: If `DBSettings` were provided on
-                instantiation and the experiment being created has the same name
-                as some experiment already stored, whether to overwrite the
-                existing experiment. Defaults to False.
+            overwrite_existing_experiment: If the experiment being created has the same
+                name as the locally created experiment or a DB stored experiment, 
+                whether to overwrite the existing experiment. Defaults to False.
             choose_generation_strategy_kwargs: Keyword arguments to pass to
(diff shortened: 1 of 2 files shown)


sdsingh:
Hey there @casassg, sorry for the delayed response here. Aside from holidays, we've had an issue with our unit tests in Travis that @lena-kashtelyan has been working to resolve. It looks like that will go in today, and then we'll be able to hopefully comment on/merge your PR :)

lena-kashtelyan:
Will merge this today, sorry for the delay!

lena-kashtelyan:
Hi, @casassg! The PR is now merged, thank you so much for it again! I had to tweak ([this portion](https://github.com/facebook/Ax/blob/master/ax/service/ax_client.py#L220-L246)) it a bit to get the tests to pass, but otherwise it's essentially what you wrote.

casassg:
That's fine, thanks for taking the time into fixing it!

PR #231, opened 2019-12-13
fix pyre for stats.norm and starts.t

xiecong:
Summary:
fix pyre error by adding
stats.norm: https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.norm.html
starts.t: https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.t.html
ppf and interval function in rv_continuous: https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.rv_continuous.html#scipy.stats.rv_continuous

Reviewed By: dkgi

Differential Revision: D18685932


sdsingh:
landed internally

Issue #237, opened 2019-12-24
ModuleNotFoundError: No module named 'botorch.models.fidelity'


ZoranPandovski:
Run a simple optimization loop throws ModuleNotFoundError, using ax-platform==0.1.6 version.
```
from ax import optimize
best_parameters, best_values, experiment, model = optimize(
        parameters=[
          {
            "name": "x1",
            "type": "range",
            "bounds": [-10.0, 10.0],
          },
          {
            "name": "x2",
            "type": "range",
            "bounds": [-10.0, 10.0],
          },
        ],
        # Booth function
        evaluation_function=lambda p: (p["x1"] + 2*p["x2"] - 7)**2 + (2*p["x1"] + p["x2"] - 5)**2,
        minimize=True,
    )
```
**Traceback**

> Traceback (most recent call last):
>     from ax.models.torch.botorch_defaults import (
>   File "/home/zoran/MyProjects/mindsdb-examples/pl/lib/python3.7/site-packages/ax/models/torch/botorch_defaults.py", line 12, in <module>
>     from botorch.models.fidelity.gp_regression_fidelity import (
> ModuleNotFoundError: No module named 'botorch.models.fidelity'

It looks like there are changes in the latest botorch 0.2.0 version. Downgrading to 0.1.4 fixes the issue.

shichangzhang-coder:
I also encountered the same problem, hope to be resolved, thanks!

shichangzhang-coder:
Downgrading to 0.1.4 can fixes the issue, thank you ZoranPandovski ^^

Balandat:
Thanks for flagging this. There were some hiccups in the tandem release we had planned for Ax & botorch, for now you can either use the latest Ax release with botorch 0.1.4 or the current master with botorh 0.2.0. 

We're currently working on cutting a new Ax release compatible with 0.2.0 that should be out soon.

huazuhao:
1) What do you mean by current master with botorch 0.2.0? Isn't the pip version already 0.2.0?
2) How exactly do we downgrade to 0.1.4? Should we just checkout a particular historical commit?

huazuhao:
Also, we tried to install the bleeding edge version of Ax but got the error code of 128 during the installation process. Thank you.

Balandat:
1. The botorch pip version is already 0.2.0, correct. You can use that with the Ax master - you can install the Ax master via pip using `pip install git+ssh://git@github.com/facebook/Ax.git` as described in the readme. 

[...]

PR #238, opened 2019-12-25
Update botorch version

ZoranPandovski:
This PR updates the botorch version to 0.1.4 instead of the latest avaiable. In the current implementation latest botorch version is installed as a dependency and #237 issue is happening. Note: Travis will fail as the latest version from master.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@ from setuptools import find_packages, setup
 
 
 REQUIRES = [
-    "botorch>=0.1.3",
+    "botorch==0.1.4",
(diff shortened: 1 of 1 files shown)


ZoranPandovski:
@Balandat I think the current master version for botorch is setup as >=0.1.3.

sdsingh:
@ZoranPandovski, totally correct. I also agree we should peg to a particular version. I'm currently launching a new Ax version, which will peg the BoTorch version to `0.2.0`. 

That should be done shortly (<2 hours), so we can probably skip this update.

PR #328, opened 2020-05-29
Expanding TConfig

abbas-kazerouni:
Summary: This adds the possibility of having a Dict[str,Any] as a value in TConfig

Reviewed By: lena-kashtelyan

Differential Revision: D21789863


saitcakmak:
already committed

PR #358, opened 2020-08-14
Remove ax/service/utils/storage from Sphinx, since that file was removed, and add WithDBSettingsBase

lena-kashtelyan:
Differential Revision: D23132365


saitcakmak:
already committed

PR #362, opened 2020-08-15
Updated Readme.md - Added MacOS version, links to dependencies

adithyaakrishna:
Changed Branch to patch-1 containing all the changes.

The changes were
a. To add all the respective links to python dependencies, in Line 70-74
image

b. To add links and macos version for recommendation, in Lines 120 & 164
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,11 +67,11 @@ You need Python 3.7 or later to run Ax.
 The required Python dependencies are:
 
 * [botorch](https://www.botorch.org)
-* jinja2
-* pandas
-* scipy
-* sklearn
-* plotly >=2.2.1
+* [jinja2](https://pypi.org/project/Jinja2/)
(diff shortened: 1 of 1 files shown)


ldworkin:
Thanks for this, @adithyaakrishna ! However with regards to the PyTorch recommendation, what we really meant was to direct readers to this paragraph:

> Recommendation for MacOS users: PyTorch is a required dependency of BoTorch, and can be automatically installed via pip. However, we recommend you install PyTorch manually before installing Ax, using the Anaconda package manager. Installing from Anaconda will link against MKL (a library that optimizes mathematical computation for Intel processors). This will result in up to an order-of-magnitude speed-up for Bayesian optimization, as at the moment, installing PyTorch from pip does not link against MKL.

The instructions there are a bit different than what is given by the link you have here.

PR #398, opened 2020-10-06
Fix broken raytune tutorial

dme65:
This fixes two issues with the raytune tutorial:
1. `ax` isn't the first argument expected by `AxSearch` which causes the error in section 4 (`TypeError: 'AxClient' object is not iterable`).
2. All points suggested by `ax` were Sobol points as no concurrency limit was specified, which hurt the optimization performance.  Passing in max_concurrent=3 limits ray from evaluating more than 3 trials in parallel and I believe this solves this issue.


lena-kashtelyan:
This is fantastic and looks great to me! @richardliaw, do these changes sound good to you?

PR #467, opened 2021-01-05
Fix raytune by adding tensorboard

Jakepodell:
This doesn't actually seem to fix the raytune tutorial, but it does get rid of the error that tensorboard is not imported


Jakepodell:
test cron: https://github.com/facebook/Ax/runs/1646906552?check_suite_focus=true

PR #476, opened 2021-01-21
Print all Sphinx warnings before treating them as errors

lena-kashtelyan:
Summary: Trying out a version where hopefully all the warnings will be printed at once

Differential Revision: D25885900


Balandat:
How about just using the `--keep-going` flag? 

> --keep-going
With -W option, keep going processing when getting warnings to the end of build, and sphinx-build exits with exit status 1.

PR #514, opened 2021-03-02
Fix json saving and loading in tutorial

justinkterry:
Using the logic from before produced this error for me:

```
  File "<stdin>", line 1, in <module>
NameError: name 'load' is not defined
```


ldworkin:
Thanks for this, @justinkterry !

PR #643, opened 2021-07-23
Remove Simple experiment from visualization tutorial

danielcohenlive:
Summary: Replaced with `Experiment` and moved eval function to stock metrics

Differential Revision: D29886059


danielcohenlive:
We decided on using the service api instead

PR #653, opened 2021-08-02
Fix BoTorch link in tutorials index page

trsvchn:
Hi everybody!

Thank you for the great project! I've just started to learn it (from tutorials). and noticed a wrong link for the BoTorch website. I think we have to use absolute path here to get it work.

Best


danielcohenlive:
Thank you, @trsvchn!

PR #660, opened 2021-08-11
Use absolute path for logo

danielcohenlive:
It should eventually fix this problem
image
see https://stackoverflow.com/questions/41983209/how-do-i-add-images-to-a-pypi-readme-that-works-on-github


Balandat (approved):
lgtm, let me do the same thing for botorch actually...

PR #665, opened 2021-08-16
Add missing `digits` parameter option to docstring

josalhor:
Hello again :)

I've found this small detail missing from a docstring. I think it should match the parameters of `make_experiment`.

Revising the docs, I wanted to bring to your attention that the [Service API tutorial code](https://ax.dev/tutorials/gpei_hartmann_service.html) still makes use of `objective_name` and `minimize`. I find it important given that #616 [deprecates](https://github.com/facebook/Ax/blob/main/ax/service/ax_client.py#L212) `objective_name` and `minimize`. I guess the objective (pun intended) is to deprecate single objective optimization in favor of MOO?

### DIFF
--- a/ax/service/ax_client.py
+++ b/ax/service/ax_client.py
@@ -208,6 +208,7 @@ class AxClient(WithDBSettingsBase):
                 parameters,
                 4. "is_ordered" (bool) for choice parameters, and
                 5. "is_task" (bool) for task parameters.
+                6. "digits" (int) for float-valued range parameters.
             name: Name of the experiment to be created.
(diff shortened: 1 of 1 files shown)


danielcohenlive:
@josalhor thanks for that fix!  Also, I want to assure you that we aren't planning on deprecating single objective optimization.  The objective is just to bring the arguments in line with a format that supports both single and multi objective optimization.  But if you pass in a single objective `objectives` dict, it will still run a single objective experiment.  You'll still be able to call `ax_client.get_best_parameters()` instead of looking at a pareto frontier

PR #683, opened 2021-08-30
(torchx/hpo) v0 proof-of-concept of torchx<>ax plugin

kiukchung:
Summary:
Adds ax<>torchx plugin. Implements:

1. TorchXScheduler (ax scheduler that delegates to TorchXRunner)
2. TorchXRunner (TorchXRunner that delegates to the actual `torchx.runner`)
3. AppMetrics (WIP - pull metrics from a blob store uri)

NOTE: ~~still need to address the TODOs and FIXMEs. Publishing early for RFC.~~ In the interest of XFN I'd like to land this as is with a couple of fast followups (to be done within a week) to put this in a workable state:
1. Properly handle trial results (currently we return trial index as the metric, which is nonsense)
1. Move the TorchX-Ax-Scheduler, Metric, and Runner implementations to `torchx.runtime.hpo.ax` (since these are designed for programmatic usages by the end user).
1. Author a generic main method for the simplest hpo search (would serve as an expository example for users to follow).
1. Add all the above to the torchx hpo docs page.

Reviewed By: d4l3k

Differential Revision: D30320103

### DIFF
--- a/ax/service/utils/best_point.py
+++ b/ax/service/utils/best_point.py
@@ -109,6 +109,35 @@ def get_best_from_model_predictions(
     return None
 
 
+def _get_best_row_for_scalarized_objective(
+    df: pd.DataFrame,
+    objective: ScalarizedObjective,
+) -> pd.DataFrame:
+    df = df.copy()
+    # First, add a weight column, setting 0.0 if the metric is not part
+    # of the objective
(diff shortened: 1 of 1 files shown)


kiukchung:
hmmm not sure why this diff ended up here, no need for this PR. closing

PR #707, opened 2021-10-19
add `sort_values` when encoding task

qingfeng10:
Summary:
`sort_values` is not passed when encoding trial index / task as ChoiceParameter. This will cause warning in the multi-task GP. The task features are encoded as integer and the values are sorted. Thus, let's add `sort_values=True` to remove the warning. (e.g. shown in N1249577 )

{F671397590}

Reviewed By: stevemandala

Differential Revision: D31731409


saitcakmak:
already committed

PR #724, opened 2021-11-16
fix in-sample arm visualization with a multi-task model

qingfeng10:
Summary:
### Problem

Visualization of in-sample arms (e.g. status-quo is None or out-of-design) is incorrect when using multi-task model. This is because we get raw metric values from an arbitrary observation rather the observations from the selected trial (e.g. selecting via FixedFeature) , e.g. as shown in this plot.

The parameter values of status-quo arm is None therefore we use observed raw value rather than model prediction. There are multiple trials and the model is a MTGP.  The `_get_in_sample_arms` uses the obs from latest short-run trial for status-quo while other arms use the MTGP prediction of long-run trial 0. Thus the relative effects are very off.  Please see N1350023 for more plots.
{F680051297}

### Fix
For out-of-design arms, we will check whether the selected trial index is passed via fixed_feature and use the observation from the same trial index.

P.S. another bug is found when data_selector is not None.  We need to modify the t`raining_in_design` list based on the selected obs.

Reviewed By: stevemandala

Differential Revision: D32420148

### DIFF
--- a/ax/plot/helper.py
+++ b/ax/plot/helper.py
@@ -173,8 +173,17 @@ def _get_in_sample_arms(
         - Map from arm name to parameters
     """
     observations = model.get_training_data()
+    training_in_design = model.training_in_design
     if data_selector is not None:
         observations = [obs for obs in observations if data_selector(obs)]
+        training_in_design = [
+            model.training_in_design[i]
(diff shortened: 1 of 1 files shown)


saitcakmak:
already committed

Issue #678, opened 2021-08-25
RuntimeError when it is run on gpu while parameter constraint is present


ertandemiral:
Hi Ax team,

Thanks for the quick fix the issue #676.

I am getting another error during GPEI generation step when I define parameter constraint to my problem and the device type is introduced as “cuda:0”. I added both my simple code and full backtrace below. Ax version is '0.2.1'.

```
from ax.modelbridge.generation_strategy import GenerationStrategy, GenerationStep
from ax.service.ax_client import AxClient
from ax.modelbridge.registry import Models
import torch

def objective_function(x):
    y = x["x1"]**2 + x["x2"]**2
    return {"y":(y, None)}

parameters=[{"name": "x1", "type": "range", "bounds": [0.0, 1.0]},{"name": "x2", "type": "range", "bounds": [0.0, 1.0]}]

gs = GenerationStrategy(
    steps=[
        GenerationStep(
            model=Models.SOBOL,
            num_trials=5,  
        ),
        GenerationStep(
            model=Models.GPEI,
            num_trials=-1,  
            model_kwargs = {"torch_dtype":torch.double, "torch_device":torch.device("cuda:0")},
        ),
    ]
)

exp = AxClient(generation_strategy = gs, verbose_logging = False)
exp.create_experiment(
                name = "dummy_exp",
                parameters = parameters,
                objective_name = "y",
                minimize = True,
                parameter_constraints = [" x1 + x2 <= 1.0"],
                )

for i in range(20):
    parameters, trial_index = exp.get_next_trial()
    value = objective_function(parameters)
    exp.complete_trial(trial_index = trial_index, raw_data = value)

```

```
  File "C:\Anaconda3\lib\site-packages\ax\utils\common\executils.py", line 169, in handle_exceptions_in_retries
    yield  # Perform action within the context manager.

  File "C:\Anaconda3\lib\site-packages\ax\utils\common\executils.py", line 147, in actual_wrapper
    return func(*args, **kwargs)

  File "C:\Anaconda3\lib\site-packages\ax\service\ax_client.py", line 327, in get_next_trial
    generator_run=self._gen_new_generator_run(), ttl_seconds=ttl_seconds

  File "C:\Anaconda3\lib\site-packages\ax\service\ax_client.py", line 1095, in _gen_new_generator_run
    return not_none(self.generation_strategy).gen(

  File "C:\Anaconda3\lib\site-packages\ax\modelbridge\generation_strategy.py", line 405, in gen
[...]

PR #821, opened 2022-02-24
Add torch_device option to AxClient

### LINKED ISSUE #811: Allow users to easily choose to run optimization on GPU in Service API
Currently that only seems possible by passing in a custom Generation Strategy with the appropriate kwargs. We don't want users to have to dig into the details of this if they want to run the optimization on a GPU (which is going to be a lot faster especially for our MOO algorithms).

saitcakmak:
Summary:
Implements a simple interface for specifying the device in service API. Resolves #811.

With this, one can opt to use GPU via `ax_client = AxClient(torch_device=torch.device("cuda"))`.
Note: this is ignored if a `generation_strategy` is passed in manually.

Reviewed By: lena-kashtelyan

Differential Revision: D34351219

### DIFF
--- a/ax/modelbridge/dispatch_utils.py
+++ b/ax/modelbridge/dispatch_utils.py
@@ -7,8 +7,9 @@
 import logging
 import warnings
 from math import ceil
-from typing import Dict, cast, Optional, Type, Union
+from typing import Any, Dict, cast, Optional, Type, Union
 
+import torch
 from ax.core.experiment import Experiment
 from ax.core.optimization_config import OptimizationConfig
 from ax.core.parameter import ChoiceParameter, ParameterType, RangeParameter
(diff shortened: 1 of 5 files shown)


sgbaird:
For visibility, to specify `torch_device` in a custom generation strategy:
```python
gs = GenerationStrategy(
    steps=[
        GenerationStep(
            model=Models.SOBOL,
            num_trials=5,
        ),
        GenerationStep(
            model=Models.GPEI,
            num_trials=-1,  
            model_kwargs = {"torch_device": torch.device("cuda")},
        ),
    ]
)

exp = AxClient(generation_strategy = gs)
```
https://github.com/facebook/Ax/issues/678#issue-979292175

Cross-posting this in https://github.com/facebook/Ax/issues/811

PR #844, opened 2022-03-11
AxGlobalStopping

abbas-kazerouni:
Summary: This implements the interface as well as a simple global stopping strategy to be used by Ax service and loop API's.

Reviewed By: Balandat

Differential Revision: D34691150


saitcakmak:
already committed

PR #864, opened 2022-03-23
Add missing init file so test stops yelling at us

danielcohenlive:
Summary: We don't need __init__.py's for buck, but OSS doesn't use buck targets, so we need it there.  I think that's what this test is getting at.

Differential Revision: D35078275


Balandat (approved):

saitcakmak:
diff abandoned

PR #877, opened 2022-03-25
improvement-based global stopping

abbas-kazerouni:
Summary: implements improvement based global stoppping

Reviewed By: danielrjiang, Balandat

Differential Revision: D34705629


saitcakmak:
already committed

PR #890, opened 2022-03-31
adding a safeguard for stopping strategy

abbas-kazerouni:
Summary: The stopping strategy at first call to get_next_trial() raises error since there are no compleleted trials yet to call max() function on. This fixes that.

Reviewed By: danielrjiang

Differential Revision: D35274315


saitcakmak:
already committed

PR #921, opened 2022-04-19
Saitcakmak patch 1

saitcakmak:
Testing. Will update once finalized.


danielcohenlive:
Looks like the test plan failed.

saitcakmak:
@danielcohenlive Yeah, it triggered a branch projection rule despite the "build-tutorials-with-pinned-botorch" workflow that the rule requires passing in the same run. I'll update once we decide on how to get around that.

saitcakmak:
abandoned

PR #958, opened 2022-05-11
Speed up aggregation with native pandas

mpolson64:
Summary: Naive implementation was too slow.

Differential Revision: D36283825


Balandat (approved):
Vectorize This!

PR #968, opened 2022-05-13
Fix typo in EI formula in BO documentation

kilian-hu:
Summary:
Fix small typo in the EI formula in the BO documentation which caused a parenthesis mismatch.


Balandat (approved):
Thanks!

PR #978, opened 2022-05-28
Pin protobuf version

saitcakmak:
The new `4.21.1` release of `protobuf` leads to errors while importing from `ray` and `tensorboard`. This pins the `protobuf` version as a temporary fix.


Balandat (approved):
Lgtm

Balandat:
Should this also be pinned in setup.py?

saitcakmak:
@Balandat We don't have `ray` or `tensorboard` as dependencies in `setup.py`, so I skipped it there. I think it is ok to skip since this is somewhat of a specialized use case and the error message pretty clearly states that you can pin `protobuf` to get around it.

PR #994, opened 2022-06-05
fix ListSurrogate.construct for constructing MTGP

qingfeng10:
Summary:
We see an error in CV when using ListSurrogate (a list of MTGP) in Models.BOTORCH_MODULAR. See N2054031 for the full error trace.

The error happens when loading state_dict. The reason is that a wrong task_feature is passed to MTGP (use default value = 0) and thus, the task_covar_module has a wrong shape.

The proposed the fix is
1.  ListSurrogate.construct. Pass task_feature to contruct MTGP. Also add checks to make sure that len(task_features) = 1
2. Make task_feature as required input in the MultiTaskGP.construct_inputs

Differential Revision: D36925653


saitcakmak:
already committed

PR #1024, opened 2022-07-15
Assert repository (hg) status doesn't change during test runs

danielcohenlive:
Summary:
I believe this will work better than D37865277 because it will tell us specifically which test failed.  D37865277 will tell us all tests that ran after the file was created.  The offending test might also be left out if it overrode `tearDown()`, which makes it very tough to debug!

This compares `hg status` on `setUp()` to `hg status` on `tearDown()`, asserting that `setUp()` was called, and that they are the same.

Differential Revision: D37901914


saitcakmak:
diff abandoned

PR #1033, opened 2022-07-26
Fix search space check bug for HSS.

EugenHotaj:
Summary:
In hierarchical search spaces, trials only contain a subset of
parameters from the search space.

Differential Revision: D38172892

### DIFF
--- a/ax/service/ax_client.py
+++ b/ax/service/ax_client.py
@@ -1694,7 +1694,9 @@ class AxClient(WithDBSettingsBase, BestPointMixin, InstantiationBase):
         # `check_membership` uses int and float interchangeably, which we don't
         # want here.
         for p_name, parameter in self.experiment.search_space.parameters.items():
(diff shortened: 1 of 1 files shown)


Balandat:
Hmm so unless I'm missing something here, one potential issue with this change is that now it's possible to also add parameters that are never in the search space via `attach_trial`. E.g. a simple typo in the parameter names would result in this passing in this validation - it might be caught downstream, but I'm not sure about that. Given that this is validating the parameterization, I think the proper thing to do here is (i) what you did, but (i) in addition also check that each parameter in the parameterization provided is indeed an element of the "full" search space (i.e. the union of all parameter names).

EugenHotaj:
Just closing the loop here since comments don't seem to be getting synched. This is just a bug in the extra type checking step, and the actual membership check happens a few lines above, which should catch non-existent parameters and typos.

PR #1040, opened 2022-08-02
Fix flake8 issues in OSS

mpolson64:
Summary: The linter we have running on github must have updated/changed settings or something bc OSS started failing out of the blue. Maybe we should consider disabling this (or disabling the internal linter, though that seems bad) rather than trying to keep both linters in sync. Thoughts?

Differential Revision: D38355972

### DIFF
--- a/ax/core/generator_run.py
+++ b/ax/core/generator_run.py
@@ -231,8 +231,8 @@ class GeneratorRun(SortableBase):
 
     @property
     def index(self) -> Optional[int]:
-        """The index of this generator run within a trial's list of generator run structs.
-        This field is set when the generator run is added to a trial.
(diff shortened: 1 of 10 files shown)


esantorella:
Same issue as https://github.com/pytorch/botorch/pull/1334. flake8 updated to 5.0.0 over the weekend. The new behavior seems to be correct since flake8 wasn't catching docstrings that were too long before this. Given that the new behavior is correct and the internal linter will update eventually, I would be in favor of leaving the linting setup how it is.

Balandat (approved):
> Given that the new behavior is correct and the internal linter will update eventually, I would be in favor of leaving the linting setup how it is.

Agreed.

PR #1055, opened 2022-08-09
Fix unit tests broken by upstream gpytorch changes

Balandat:
cornellius-gp/gpytorch#2082 changed the shape of the mean constant (and did some other things). These changes fix the resulting unit test breakages.

This requires the botorch change https://github.com/pytorch/botorch/pull/1343


saitcakmak (approved):
Lgtm

PR #1068, opened 2022-08-11
Support mixing environmental and perturbation parameters in `RobustSearchSpace`

saitcakmak:
Summary: Updates `RobustSearchSpace` to support having input perturbations alongside of environmental variables. Updates `RobustSearchSpaceDigest` to have two fields for distribution samplers, one for perturbation distributions and one for environmental variables, and updates `extract_robust_digest` to construct the two samplers together.

Differential Revision: D38598604

### DIFF
--- a/ax/core/parameter_distribution.py
+++ b/ax/core/parameter_distribution.py
@@ -8,11 +8,14 @@ from __future__ import annotations
 import functools
 from copy import deepcopy
 from importlib import import_module
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, TYPE_CHECKING
 
 from ax.exceptions.core import UserInputError
 from ax.utils.common.base import SortableBase
 from scipy.stats._distn_infrastructure import rv_generic
+
+if TYPE_CHECKING:
+    from ax.core.search_space import RobustSearchSpace
 
 TDistribution = str
 TParamName = str
@@ -67,6 +70,18 @@ class ParameterDistribution(SortableBase):
                 "`distribution_class` is importable from `scipy.stats`."
             )
         return dist_class(**self.distribution_parameters)
+
+    def is_environmental(self, search_space: RobustSearchSpace) -> bool:
+        r"""Check if the parameters are environmental variables of the given
+        search space.
+
+        Args:
+            search_space: The search space to check.
+
+        Returns:
+            A boolean denoting whether the parameters are environmental variables.
+        """
+        return any(search_space.is_environmental_variable(p) for p in self.parameters)
 
     def clone(self) -> ParameterDistribution:
         """Clone."""
--- a/ax/core/search_space.py
+++ b/ax/core/search_space.py
@@ -697,38 +697,15 @@ class RobustSearchSpace(SearchSpace):
                 "RobustSearchSpace requires at least one distributional parameter. "
                 "Use SearchSpace instead."
             )
-        if num_samples < 1:
+        if num_samples < 1 or int(num_samples) != num_samples:
             raise UserInputError("`num_samples` must be a positive integer!")
         self.num_samples = num_samples
-        # Make sure that the distributions are all multiplicative or additive.
-        mul_flags = [d.multiplicative for d in parameter_distributions]
-        if not (all(mul_flags) or not any(mul_flags)):
-            raise UnsupportedError(
-                "Parameter distributions must be either all multiplicative "
-                "or all additive (not multiplicative)."
-            )
         self.parameter_distributions = parameter_distributions
-        # Make sure that there is at most one distribution per parameter.
-        distributional_parameters_list = []
-        for param_dist in parameter_distributions:
-            distributional_parameters_list.extend(param_dist.parameters)
-        self._distributional_parameters: Set[str] = set(distributional_parameters_list)
-        if len(self._distributional_parameters) != len(distributional_parameters_list):
-            raise UserInputError(
-                "Received multiple parameter distributions for at least one parameter. "
-                "Make sure that there is at most one distribution specified for any "
-                "given parameter / environmental variable."
-            )
-        # Make sure that all env vars have distributions specified,
-        # and the names are unique.
+        # Make sure that the env var names are unique.
         environmental_variables = environmental_variables or []
         all_env_vars: Set[str] = {p.name for p in environmental_variables}
         if len(all_env_vars) < len(environmental_variables):
             raise UserInputError("Environmental variable names must be unique!")
-        if not all_env_vars.issubset(self._distributional_parameters):
-            raise UserInputError(
-                "All environmental variables must have a distribution specified."
-            )
         self._environmental_variables: Dict[str, Parameter] = {
             p.name: p for p in environmental_variables
         }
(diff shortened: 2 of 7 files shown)


ax/core/parameter_distribution.py
```diff
@@ -68,6 +71,21 @@ def distribution(self) -> rv_generic:
             )
         return dist_class(**self.distribution_parameters)
 
+    def is_environmental(self, search_space: RobustSearchSpace) -> bool:
+        r"""Check if the parameters are environmental variables of the given
+        search space.
+
+        It is sufficient to check the first parameter since the `RobustSearchSpace`
+        does not allow mixing the two types of distributions.
```
Balandat, line 79: Should we still check with `any` here to be future-proof in case we change that down the road?

ax/core/search_space.py
```diff
@@ -744,37 +721,95 @@ def __init__(
...
+        dist_params_list = []
+        for dist in distributions:
+            dist_params_list.extend(dist.parameters)
+        self._distributional_parameters: Set[str] = set(dist_params_list)
+        if len(self._distributional_parameters) != len(dist_params_list):
+            raise UserInputError(
+                "Received multiple parameter distributions for at least one parameter. "
+                "Make sure that there is at most one distribution specified for any "
+                "given parameter / environmental variable."
+            )
```
Balandat, line 746: Can raise a more informative error here
```suggestion
        self._distributional_parameters: Set[str] = set()
        for dist in distributions:
            duplicates = self._distributional_parameters.intersection(dist.parameters)
            if duplicates:
                raise UserInputError(
                    f"Received multiple parameter distributions for parameters {duplicates}. "
                    "Make sure that there is at most one distribution specified for any "
                    "given parameter / environmental variable."
                )
            self._distributional_parameters.update(dist.parameters)
```

ax/modelbridge/modelbridge_utils.py
```diff
@@ -189,61 +192,70 @@ def extract_robust_digest(
...
             raise RuntimeError(
                 "Environmental variables must be last entries in `param_names`. "
                 "Otherwise, `AppendFeatures` will not work."
             )
         # NOTE: Extracting it from `param_names` in case the ordering is different.
-        environmental_variables = param_names[num_non_dist_params:]
+        environmental_variables = param_names[num_non_env_vars:]
 
-        def get_samples() -> np.ndarray:
+        def environmental_sampler() -> np.ndarray:
```
Balandat, line 207: Semantic nit
```suggestion
        def ample_environmental() -> np.ndarray:
```

ax/modelbridge/modelbridge_utils.py
```diff
@@ -189,61 +192,70 @@ def extract_robust_digest(
...
 
     else:
+        environmental_sampler = None
         environmental_variables = []
 
-        def get_samples() -> np.ndarray:
+    if len(pert_params) > 0:
+        constructor = np.ones if search_space.multiplicative else np.zeros
+
+        def distribution_sampler() -> np.ndarray:
```
Balandat, line 232: "distribution" as a term is a bit unclear - we're always sampling from distributions. Assuming this is about (possibly degenerate) parameter perturbations how about the following (assuming this doesn't cause a bunch of backward compatibility issues if we also change it in the `RobustSearchSpace`)?
```suggestion
        def sample_param_perturbations() -> np.ndarray:
```

ax/modelbridge/transforms/unit_x.py
```diff
@@ -109,102 +110,118 @@ def untransform_observation_features(
...
-                / p_range
-                * self.target_range
+                self._transform_univariate_distribution(
+                    distribution=distribution, is_environmental=is_environmental
+                )
+
+    def _transform_univariate_distribution(
+        self, distribution: ParameterDistribution, is_environmental: bool
+    ) -> None:
+        r"""Transform a univariate distribution in-place."""
```
Balandat, line 139: Is there a reason to do this in-place rather than return a new object?

ax/modelbridge/transforms/unit_x.py
```diff
@@ -109,102 +110,118 @@ def untransform_observation_features(
...
-                    "parameter and the corresponding distribution."
+
+    def _transform_multivariate_distribution(
+        self, distribution: ParameterDistribution, is_environmental: bool
+    ) -> None:
+        r"""Transform a multivariate distribution in-place."""
+        # Ignore if the ranges of all parameters are same as the target range.
+        if (
+            all(
+                self.bounds[p_name][1] - self.bounds[p_name][0] == self.target_range
```
Balandat, line 176:
```suggestion
                np.diff(self.bounds[p_name]) == self.target_range
```

ax/modelbridge/transforms/unit_x.py
```diff
@@ -109,102 +110,118 @@ def untransform_observation_features(
...
+            cov = np.asarray(distribution.distribution_parameters.get("cov"))
+            if cov.shape != (n_dist_params, n_dist_params):
+                raise UserInputError(
+                    "Expected `cov` to be a square matrix of size equal to "
+                    "number of parameters. Received `cov` with shape "
+                    f"{cov.shape} for the distribution of {n_dist_params} "
+                    "parameters."
                 )
+            distribution.distribution_parameters["cov"] = (
+                scale_vec[..., None] * cov * scale_vec[..., None, :]
```
Balandat, line 200:
```suggestion
                # same as np.diag(scale_vec) @ cov @ np.diag(scale_vec) but faster
                scale_vec[..., None] * cov * scale_vec[..., None, :]
```

ax/modelbridge/transforms/unit_x.py
```diff
@@ -109,102 +110,118 @@ def untransform_observation_features(
...
+            )
+            mean = np.asarray(distribution.distribution_parameters.get("mean", 0.0))
+            if not np.all(mean == 0):
+                if mean.shape != (n_dist_params,):
```
Balandat, line 204: Do we not want to throw the shape error if mean is not of the right shape? Or do we also want to allow passing scalars? I think we can be pretty restrictive in what we support here as it's deep down in the code (if we wanted to support user friendly interfaces we could do that further up in the stack).

Balandat (approved):
Thanks! Some comments on naming, o/w these are all nits.

ax/modelbridge/modelbridge_utils.py
saitcakmak: yeah, I considered renaming this to `perturbation_sampler` but left it as is mostly due to laziness. This stuff is new enough and is still not supported throughout the modeling stack (except for my stack of diffs), so it should be ok to break BC.

ax/modelbridge/transforms/unit_x.py
saitcakmak: The general pattern in Ax transforms is to do the transforms in-place. So, I just went with that. I think it also leads to a simpler implementation.

ax/modelbridge/transforms/unit_x.py
saitcakmak: I'd generally expect the non-zero mean case to be limited to environmental variables only. Since this is an advanced feature anyway, I prefer to be restrictive and require the user to be explicit in what they want to achieve.

Padarn:
Hi @saitcakmak I'm sorry to come to your PR randomly. I'm trying to understand the reason for requiring that all of the perturbation distributions are either multiplicative or additive?

Do you have any design docs for this new set of features I can use as a reference?

saitcakmak:
Hi @Padarn. The main reason for requiring them to all be multiplicative or additive is that it simplifies implementation. We do not have any current use cases for multiplicative perturbations (other than the synthetic examples we had in the paper), so they have a bit more limited support (e.g. they don't support transforms, so your search space has to be the unit hypercube). Do you have a particular use case that requires mixing additive and multiplicative transforms? 

> Do you have any design docs for this new set of features I can use as a reference?

We have some tutorials in BoTorch (e.g. https://botorch.org/tutorials/robust_multi_objective_bo) but I haven't written any for Ax. I have some internal design docs that I wrote at various points while implementing this. I can isolate the OSS parts and share that with you.

Padarn:
Yeah, if you don't mind, that would be cool. The `botorch` example helps a lot already though thanks!
> Do you have a particular use case that requires mixing additive and multiplicative transforms?
No I don't, I was just trying to understand the implementation and couldn't quite understand the reason for making the restriction so thought I must be missing something obvious.

saitcakmak:
@Padarn Here's a (very rough) design doc that covers a decent part of the implementation. There are things that are not included in there such as the support for (non-environmental) robust optimization in modular BoTorch model (which is the only model that supports robust opt out of box). Hope it helps!
[Ax robust opt.rtf.zip](https://github.com/facebook/Ax/files/10126303/Ax.robust.opt.rtf.zip)

Padarn:
Sorry for the slow response, this is super useful, thanks for taking the time to make it sharable.

PR #1071, opened 2022-08-15
Bump BoTorch version prior to release

mpolson64:
Summary: Current Ax version is using botorch 0.6.4, causing errors. Bumping required version of botorch to the current botorc release. An Ax release is to be made after this commit.

Differential Revision: D38711411


Balandat (approved):
bump it!

PR #1113, opened 2022-09-06
Deterministic order of TorchModelBridge paramaters in _fit

ItsMrLin:
Summary: Make fitted TorchModelBridge parameters in a particular order.

Reviewed By: qingfeng10

Differential Revision: D39290841


Balandat:
Can we add a check for this in the unit test_

ItsMrLin:
Closed this PR since we realized instead of simple fix, this would impact many other code/tests. Will proceed with a different solution on the metric ordering issue to support BOPE.

PR #1120, opened 2022-09-09
Support loading a state dict for SaasFullyBayesianSingleTaskGP (#1384)

dme65:
Summary:
X-link: https://github.com/pytorch/botorch/pull/1384

`SaasFullyBayesianSingleTaskGP` currently doesn't support `load_state_dict` since the model is only initialized after fitting, so we can't load a state dict into a model that hasn't been fitted. This diff modifies `load_state_dict` to initialize the model with some dummy samples before loading the state dict.

Reviewed By: saitcakmak, Balandat

Differential Revision: D39358160


dme65:
Lint and test errors are unrelated

PR #1121, opened 2022-09-09
early stopping tutorial

danielrjiang:
Summary: es tutorial

Differential Revision: D39376244


tutorials/early_stopping/early_stopping.ipynb
```diff
@@ -0,0 +1,1067 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "id": "0775a96e",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "class MyTensorboardMetric(TensorboardCurveMetric):\n",
```
Balandat, line 231: I wonder if it makes sense to check in sth like a `LocalDirTensorBoardCurveMetric` that can be instantiated with the log dir so that this can just be imported rather than needing to be defined.

tutorials/early_stopping/early_stopping.ipynb
```diff
@@ -0,0 +1,1067 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "d4f3ba5d",
+   "metadata": {},
+   "source": [
+    "## Setting up the OptimizationConfig\n",
+    "\n",
+    "The `OptimizationConfig` specifies the objective for Ax to optimize."
```
Balandat, line 273: Probably good to point to the fact that Ax also supports constraints and/or MOO. Though that might raise some questions about ES for MOO that we may want to work out first...

tutorials/early_stopping/early_stopping.ipynb
```diff
@@ -0,0 +1,1067 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "57aa9cf7",
+   "metadata": {},
+   "source": [
+    "## Defining an Early Stopping Strategy\n",
+    "\n",
+    "A `PercentileEarlyStoppingStrategy` is a simple method that stops a trial if its performance falls below a certain percentile of other trials at the same step (e.g., when `percentile_threshold` is 50, at a given point in time, if a trial ranks in the bottom 50% of trials, it is stopped). \n",
```
Balandat, line 298: Before jumping in to the specific strategy used here, could you take a step back and describe `EarlyStoppingStrategies` more generally and how they work / what their interface looks like?

tutorials/early_stopping/early_stopping.ipynb
```diff
@@ -0,0 +1,1067 @@
...
+    "- We make use of `normalize_progressions` which normalizes the progression column (e.g. timestamp, epochs, training data used) to be in [0, 1]. This is useful because one doesn't need to know the maximum progression values of the curve (which might be, e.g., the total number of data points in the training dataset).\n",
```
Balandat, line 299: we should be very clear / explicit about what the "progression" is in the context of this tutorial.

tutorials/early_stopping/early_stopping.ipynb
```diff
@@ -0,0 +1,1067 @@
...
+    "- The `min_progression` parameter specifies that trials should only be considered for stopping if the latest progression value is greater than this threshold.\n",
+    "- The `min_curves` parameter specifies the minimum number of completed curves before early stopping will be considered. This should be larger than zero if `normalize_progression` is used. In general, we want a few completed curves to have a baseline for comparison."
```
Balandat, line 301:
```suggestion
    "- The `min_curves` parameter specifies the minimum number of completed curves (i.e. fully run training jobs) before early stopping will be considered. This should be larger than zero if `normalize_progression` is used. In general, we want a few completed curves to have a baseline for comparison."
```

tutorials/early_stopping/early_stopping.ipynb
```diff
@@ -0,0 +1,1067 @@
...
+    "    # stop if in bottom 50% of runs at the same progression\n",
+    "    percentile_threshold=50,\n",
+    "    # the trial must have passed `min_progression` steps before early stopping is initiated\n",
+    "    # note that we are using `normalize_progressions`, so this is on a scale of [0, 1]\n",
+    "    min_progression=0.3,\n",
+    "    # there must be `min_curves` completed trials and `min_curves` trials reporting data in\n",
+    "    # order for early stopping to be applicable\n",
+    "    min_curves=3,\n",
+    "    # specify, e.g., [0, 1] if the first two trials should never be stopped\n",
+    "    trial_indices_to_ignore=None,\n",
```
Balandat, line 321: I don' think we need to talk about this if we don't use it in the tutorial - less is more; folks will always find this in the docstring
```suggestion
```

tutorials/early_stopping/early_stopping.ipynb
```diff
@@ -0,0 +1,1067 @@
...
+    "A [GenerationStrategy](https://ax.dev/api/modelbridge.html#ax.modelbridge.generation_strategy.GenerationStrategy)_\n",
+    "is the abstract representation of how we would like to perform the\n",
+    "optimization. While this can be customized (if you’d like to do so, see\n",
+    "[this tutorial](https://ax.dev/tutorials/generation_strategy.html)_),\n",
+    "in most cases Ax can automatically determine an appropriate strategy\n",
+    "based on the search space, optimization config, and the total number\n",
+    "of trials we want to run.\n",
+    "\n",
+    "Typically, Ax chooses to evaluate a number of random configurations\n",
+    "before starting a model-based Bayesian Optimization strategy."
```
Balandat, line 369: We should probably talk about (or at least hint at) the fact that generation strategy and early stopping strategy are interdependent, and that in an ideal world we'd consider this as one big problem, but that this generally works quite well in practice and is a lot simpler to set up.

tutorials/early_stopping/early_stopping.ipynb
```diff
@@ -0,0 +1,1067 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "d6ec43d9",
+   "metadata": {},
+   "source": [
+    "## Results\n",
+    "\n",
+    "Below is a summary of the experiment, showing that a portion of trials have been early stopped."
```
Balandat, line 457: the df right after this doesn't actually show that trials were early stopped (that is shown by the trial status in the next one). Maybe move the first one one cell back and add a comment on how that df shows the raw data of the curve observations

tutorials/early_stopping/early_stopping.ipynb
```diff
@@ -0,0 +1,1067 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "ddc543de",
+   "metadata": {},
+   "source": [
+    "## Visualizations\n",
+    "\n",
+    "Finally, we show a visualization of learning curves versus actual elapsed wall time. This helps to illustrate that stopped trials make room for additional trials to be run."
```
Balandat, line 971: Maybe ccomment here on the fact that things are run in a fully async fashion with 3 trials running concurrently (and the gaps is where the candidate generation happens)>

Balandat: It would also be nice to discuss some of the key points here, e.g. 
- the fact that we're getting better over time, 
- that really bad trials are stopped very quickly , while others may run for longer
- what the tl;dr is and why we want this (saving compute and wall time; in general it would be nice to give an estimates of the savings in this example compared to the baseline of running everything until the end (for the purposes of this tutorial we can probably assume that we identify the same best performing trial that we would have had we run the full training for all trials.

PR #1132, opened 2022-09-14
Fix constraint bug in `infer_reference_point_from_experiment`

saitcakmak:
Summary: Prior to this change, `multiplier` would only include values for the objectives and `f` would include both the objectives and the constrains, leading to a shape error.

Differential Revision: D39491606

### DIFF
--- a/ax/plot/pareto_utils.py
+++ b/ax/plot/pareto_utils.py
@@ -519,7 +519,6 @@ def infer_reference_point_from_experiment(
     Returns:
         List of obejective thresholds representing the reference point.
     """
-
     if not experiment.is_moo_problem:
         raise ValueError(
             "This function works for MOO experiments only."
@@ -538,7 +537,7 @@ def infer_reference_point_from_experiment(
         multiplier.append(-1 if objective.minimize else +1)
 
     # A dummy reference point so that all observed points are considered when
-    # calculating the parto front.
+    # calculating the Pareto front.
     dummy_rp = copy.deepcopy(
         experiment.optimization_config.objective_thresholds  # pyre-ignore
     )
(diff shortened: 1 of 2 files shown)


ax/plot/pareto_utils.py
```diff
@@ -555,7 +554,10 @@ def infer_reference_point_from_experiment(
     )
 
     # Transforming all the objectives to be maximized.
-    f_transformed = torch.tensor(multiplier, dtype=f.dtype, device=f.device) * f
+    f_transformed = (
+        torch.tensor(multiplier, dtype=f.dtype, device=f.device)
+        * f[:, obj_w.nonzero().view(-1)]
```
Balandat, line 559: Shouldn't the `view(-1)` should be a noop here?

Balandat (approved):

ax/plot/pareto_utils.py
saitcakmak: `obj_w.nonzero()` returns an `m x 1` tensor. Without the `view` the result ends up with an extra dimension.

Balandat: Looks like the docstring of `get_pareto_frontier_and_configs` is incorrect then? https://github.com/saitcakmak/Ax/blob/7e7b178e67aa7a1b061fc3dc34bf00000b078ed6/ax/modelbridge/modelbridge_utils.py#L798

saitcakmak: It's just the behavior of nonzero. `torch.rand(3).nonzero()` returns `tensor([[0], [1], [2]])`. If you use `as_tuple=True`, then it returns `(tensor([0, 1, 2]),)`. So, you either need to use the view to flatten the default or index into the tuple.

Balandat: Ah I see, thanks.

PR #1138, opened 2022-09-14
Clean up workflow pip installs, run tests on Py3.10

saitcakmak:
Summary: As titled. The # of `pip install` statements were getting out of hand...

Differential Revision: D39519257


Balandat (approved):
gtmtl, thanks

PR #1146, opened 2022-09-15
Add annotation to logger

mpolson64:
Summary:
Address one of the easiest and most common (~250!) pyre-fixmes in our codebase.

```
# pyre-fixme[5]: Global expression must be annotated.
logger = get_logger(__name__)
```
becomes
```
from logger import Logger
logger: Logger = get_logger(__name__)
```
then arc lint is applied.

Differential Revision: D39547946


Balandat (approved):
yay!

Issue #1183, opened 2022-10-05
I can't import from ax library. (OS error: Failed to open file)


CML-DHK:
I already download ax library from ax homepage. But, I can't import anything from ax library.

Ex.) from ax.service.ax_client import AxClient

OSError                                   Traceback (most recent call last)
Cell In [1], line 2
      1 from typing import Any, Dict, Optional, Tuple, Type
----> 2 from ax.service.ax_client import AxClient
      4 # Ax wrappers for BoTorch components
      5 from ax.models.torch.botorch_modular.model import BoTorchModel

File C:\Anaconda\envs\botorch\lib\site-packages\ax\__init__.py:7
      1 #!/usr/bin/env python3
      2 # Copyright (c) Meta Platforms, Inc. and affiliates.
      3 #
      4 # This source code is licensed under the MIT license found in the
      5 # LICENSE file in the root directory of this source tree.
----> 7 from ax.core import (
      8     Arm,
      9     BatchTrial,
     10     ChoiceParameter,
     11     ComparisonOp,
     12     Data,
     13     Experiment,
     14     FixedParameter,
     15     GeneratorRun,
     16     Metric,
     17     MultiObjective,
     18     MultiObjectiveOptimizationConfig,
     19     Objective,
     20     ObjectiveThreshold,
     21     OptimizationConfig,
     22     OrderConstraint,
     23     OutcomeConstraint,
     24     Parameter,
     25     ParameterConstraint,
     26     ParameterType,
     27     RangeParameter,
     28     Runner,
     29     SearchSpace,
     30     SumConstraint,
     31     Trial,
     32 )
     33 from ax.modelbridge import Models
     34 from ax.service import OptimizationLoop, optimize

File C:\Anaconda\envs\botorch\lib\site-packages\ax\core\__init__.py:9
      1 #!/usr/bin/env python3
      2 # Copyright (c) Meta Platforms, Inc. and affiliates.
      3 #
   (...)
      6 
      7 # flake8: noqa F401
      8 from ax.core.arm import Arm
----> 9 from ax.core.batch_trial import BatchTrial
     10 from ax.core.data import Data
     11 from ax.core.experiment import Experiment

File C:\Anaconda\envs\botorch\lib\site-packages\ax\core\batch_trial.py:27
     25 import numpy as np
     26 from ax.core.arm import Arm
---> 27 from ax.core.base_trial import BaseTrial
[...]

PR #1186, opened 2022-10-05
Update imports & caching in `ParameterDistribution`

saitcakmak:
Summary:
Replaces the scipy `rv_generic` import (from a private file) with a `rv_frozen` import (from a public file), and clears up the use of `lru_cache` (flake8 doesn't like it).

Additional context:  https://github.com/facebook/Ax/issues/1183#issuecomment-1268517437

Differential Revision: D40112537

### DIFF
--- a/ax/core/parameter_distribution.py
+++ b/ax/core/parameter_distribution.py
@@ -5,17 +5,16 @@
 
 from __future__ import annotations
 
-import functools
 from copy import deepcopy
 from importlib import import_module
 from typing import Any, Dict, List, Optional, TYPE_CHECKING
 
 from ax.exceptions.core import UserInputError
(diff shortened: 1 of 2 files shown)


Balandat:
> clears up the use of lru_cache (flake8 doesn't like it)

What exactly is the issue here? In general it would be good if we could use `lru_cache` without flake8 being the limiting factor.

Balandat (approved):

saitcakmak:
flake8 has a warning about `lru_cache` potentially leading to memory leaks. 

> B019: Use of functools.lru_cache or functools.cache on class methods can lead to memory leaks. The cache may retain instance references, preventing garbage collection.

PR #1187, opened 2022-10-06
Better default for wf_run_title_prefix

bernardbeckerman:
Summary: Makes SimpleInputs default to "Evaluation of arm " for wf_run_title_prefix.

Differential Revision: D40114178


saitcakmak:
merged

PR #1223, opened 2022-10-23
es tutorial

danielrjiang:
Summary: early stopping tutorial

Differential Revision: D40624123


tutorials/early_stopping/mnist_train_nas.py
```diff
@@ -0,0 +1,181 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+Example training code, adapted from
+https://pytorch.org/tutorials/intermediate/ax_multiobjective_nas_tutorial.html
```
Balandat, line 8: The tutorials are BDS3, so if we write this was adopted from there then technically we'd have to restate that license / its conditions here. Since we're the ones who wrote that code in the first place we don't need to, so maybe let's just remove this comment?

tutorials/early_stopping/early_stopping.ipynb
```diff
@@ -0,0 +1,978 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "8158a8f5",
+   "metadata": {},
+   "source": [
+    "## Early stopping in Ax\n",
+    "\n",
+    "This tutorial illustrates how to add an early stopping strategy to an Ax HPO loop. Most of this tutorial is adapted from the [PyTorch Ax Multiobjective NAS Tutorial](https://pytorch.org/tutorials/intermediate/ax_multiobjective_nas_tutorial.html). The training job is different from the original in that we do not optimize `batch_size` or `epochs`. This was done for illustrative purposes, as each validation curve now has the same number of points. The companion training file from the NAS tutorial `mnist_train_nas.py` has also been altered to log to Tensorboard during training.\n",
+    "\n",
+    "NOTE: Although the original NAS tutorial is for a multi-objective problem, this tutorial focuses on a single objective (validation accuracy) problem. Early stopping currently does not support \"true\" multi-objective stopping, although one can use [logical compositions of early stopping strategies](https://github.com/facebook/Ax/blob/main/ax/early_stopping/strategies/logical.py) to target multiple objectives separately."
```
Balandat, line 12: Might make sense to indicate that ES for MOO is in the works?

tutorials/early_stopping/early_stopping.ipynb
```diff
@@ -0,0 +1,978 @@
...
+   "cell_type": "markdown",
+   "id": "65f7011d",
+   "metadata": {},
+   "source": [
+    "## Setting up the Runner\n",
+    "\n",
+    "Ax’s [Runner](https://ax.dev/api/core.html#ax.core.runner.Runner)_\n",
+    "abstraction allows writing interfaces to various backends.\n",
+    "Ax already comes with Runner for TorchX, and so we just need to\n",
+    "configure it. For the purpose of this tutorial we run jobs locally\n",
```
Balandat, line 115: May make sense here to give a heads up that the training job will be logging partial results to tensorboard, which we will read while the training is ongoing in order to implement trial-level early stopping.

tutorials/early_stopping/early_stopping.ipynb
```diff
@@ -0,0 +1,978 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "8158a8f5",
+   "metadata": {},
+   "source": [
+    "## Early stopping in Ax\n",
+    "\n",
+    "This tutorial illustrates how to add an early stopping strategy to an Ax HPO loop. Most of this tutorial is adapted from the [PyTorch Ax Multiobjective NAS Tutorial](https://pytorch.org/tutorials/intermediate/ax_multiobjective_nas_tutorial.html). The training job is different from the original in that we do not optimize `batch_size` or `epochs`. This was done for illustrative purposes, as each validation curve now has the same number of points. The companion training file from the NAS tutorial `mnist_train_nas.py` has also been altered to log to Tensorboard during training.\n",
```
Balandat, line 10: Let's talks about "trial-level early stopping" to be precise and distinguish this from experiment-level early stopping. Maybe it's good to add a high-level discussion how trial-level ES works and why one may want to use it?

tutorials/early_stopping/early_stopping.ipynb
```diff
@@ -0,0 +1,978 @@
...
+   "id": "57aa9cf7",
+   "metadata": {},
+   "source": [
+    "## Defining an Early Stopping Strategy\n",
+    "\n",
+    "A `PercentileEarlyStoppingStrategy` is a simple method that stops a trial if its performance falls below a certain percentile of other trials at the same step (e.g., when `percentile_threshold` is 50, at a given point in time, if a trial ranks in the bottom 50% of trials, it is stopped). \n",
+    "- We make use of `normalize_progressions` which normalizes the progression column (e.g. timestamp, epochs, training data used) to be in [0, 1]. This is useful because one doesn't need to know the maximum progression values of the curve (which might be, e.g., the total number of data points in the training dataset).\n",
+    "- The `min_progression` parameter specifies that trials should only be considered for stopping if the latest progression value is greater than this threshold.\n",
+    "- The `min_curves` parameter specifies the minimum number of completed curves before early stopping will be considered. This should be larger than zero if `normalize_progression` is used. In general, we want a few completed curves to have a baseline for comparison."
```
Balandat, line 296:
```suggestion
    "- The `min_curves` parameter specifies the minimum number of completed curves (i.e., fully completed training jobs) before early stopping will be considered. This should be larger than zero if `normalize_progression` is used. In general, we want a few completed curves to have a baseline for comparison."
```

tutorials/early_stopping/early_stopping.ipynb
```diff
@@ -0,0 +1,978 @@
...
+   "cell_type": "markdown",
+   "id": "49a4ed0e",
+   "metadata": {},
+   "source": [
+    "## Choosing the GenerationStrategy\n",
+    "\n",
+    "A [GenerationStrategy](https://ax.dev/api/modelbridge.html#ax.modelbridge.generation_strategy.GenerationStrategy)_\n",
+    "is the abstract representation of how we would like to perform the\n",
+    "optimization. While this can be customized (if you’d like to do so, see\n",
+    "[this tutorial](https://ax.dev/tutorials/generation_strategy.html)_),\n",
```
Balandat, line 358: There is an excess underscore in the rendered tutorial somehow

tutorials/early_stopping/early_stopping.ipynb
```diff
@@ -0,0 +1,978 @@
...
+    "    runner=ax_runner,\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "49a4ed0e",
+   "metadata": {},
+   "source": [
+    "## Choosing the GenerationStrategy\n",
```
Balandat, line 353: It may make sense to acknowledge here that "proper" decision theoretic handling would treat generation and early stopping strategy jointly in one strategy, and that the separation here is for simplicity (and b/c it generally works quite well in practice).

tutorials/early_stopping/early_stopping.ipynb
```diff
@@ -0,0 +1,978 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "57aa9cf7",
+   "metadata": {},
+   "source": [
+    "## Defining an Early Stopping Strategy\n",
+    "\n",
+    "A `PercentileEarlyStoppingStrategy` is a simple method that stops a trial if its performance falls below a certain percentile of other trials at the same step (e.g., when `percentile_threshold` is 50, at a given point in time, if a trial ranks in the bottom 50% of trials, it is stopped). \n",
```
Balandat, line 293: Should we mention model-based ES strategies here or is that a bit too much preview / information overload?

tutorials/early_stopping/early_stopping.ipynb
```diff
@@ -0,0 +1,978 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "bc2af4c6",
+   "metadata": {},
+   "source": [
+    "## Results\n",
+    "\n",
+    "Below is a summary of the experiment, showing that a portion of trials have been early stopped."
```
Balandat, line 452: This is a bit confusing since the next df is about the progression and doesn't contain any info about ES. Suggest adding a comment on the next df here and moving the discussion for the ES trials into the next cell.

tutorials/early_stopping/early_stopping.ipynb
```diff
@@ -0,0 +1,978 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "a06172ba",
+   "metadata": {},
+   "source": [
+    "## Visualizations\n",
+    "\n",
+    "Finally, we show a visualization of learning curves versus actual elapsed wall time. This helps to illustrate that stopped trials make room for additional trials to be run."
```
Balandat, line 883: Maybe expand / elaborate a bit on this figure? 

I also wonder if there is a rough back of the envelope calculation we could do here with the result form this run to estimate what kinds of savings one might expect. A simple thing would be to make the assumption that the ES trials wouldn't have been the best and would also not have substantially affected the model, then we can give a rough ball park number for how much compute / wall time this would have saved relative to the full run (acknowledging the obvious flaws with this estimate).

PR #1230, opened 2022-10-29
Add missing parenthesis in boolean condition

### LINKED ISSUE #1225: `AxClient.get_best_trial(use_model_predictions=False)` never finds point satisfying outcome constraints
This problem seems to be solved when replacing 
https://github.com/facebook/Ax/blob/b0fe0c52179e417e1b34b55131a23d9297afd4e4/ax/service/utils/best_point.py#L697
by
```
 return ~name_match_mask | (observed_upper_bound > oc.bound)
```
and
https://github.com/facebook/Ax/blob/b0fe0c52179e417e1b34b55131a23d9297afd4e4/ax/service/utils/best_point.py#L699
by
```
 return ~name_match_mask | (observed_lower_bound < oc.bound)
```

olivierkeke:
Resolves #1225

### DIFF
--- a/ax/service/utils/best_point.py
+++ b/ax/service/utils/best_point.py
@@ -694,9 +694,9 @@ def _filter_feasible_rows(
         # Return True if metrics are different, or whether the confidence
         # interval is entirely not within the bound
         if oc.op == ComparisonOp.GEQ:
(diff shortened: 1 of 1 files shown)


Balandat:
This is addressed in #1218 - but I guess we can get this in right away

PR #1237, opened 2022-11-02
Allow loading of PEXs

danielcohenlive:
Summary:
Fixes https://www.internalfb.com/intern/anp/view/?id=2750501&checkpoint_id=1318722322235905&source=version_selector.  Previously the client tried to initiate PEX's runner with _from_client, which wasn't a valid arg, and exploded!

Fixed in 2 ways:
1. `LooperVMRunner.__init__` takes `_from_client`.  If we didn't do this it'd still be warning that you're not using PTSClient.
2.  We filter kwargs in the decoder.  This will unblock us if we want to use PTSClient for any new runners

Differential Revision: D40952199


saitcakmak:
merged

PR #1258, opened 2022-11-14
remove check that objective_thresholds and objectives are the same length

bernardbeckerman:
Summary:
--

Created from CodeHub with https://fburl.com/edit-in-codehub

Reviewed By: saitcakmak

Differential Revision: D41037113


saitcakmak:
merged

PR #1290, opened 2022-11-28
Deleting third-party-source/native/pyro

kittipatv:
Summary: Prepare to upgrade pyro via pypi. See https://fb.workplace.com/groups/402454550753394/posts/890405038625007 for context

Differential Revision: D41559889


saitcakmak:
merged

PR #1295, opened 2022-11-29
Turnoff buggy lazy tensor computations in benchmarks

saitcakmak:
Summary: This ensures that we can run benchmarks for many iterations without running into lazy tensor related bugs with SAAS models.

Reviewed By: Balandat

Differential Revision: D41569569


saitcakmak:
already committed

PR #1309, opened 2022-12-08
Delete data on experiment for trials marked bad

mpolson64:
Summary: Prevent experiment from getting into a state where a trial is completed but some critical metrics have failed to fetch, which causes errors downstream. This will not delete the data from a trial if only tracking metrics have failed to fetch, or if an available while running metric has failed to fetch while the experiment was running

Differential Revision: D41856845


saitcakmak:
abandoned

PR #1321, opened 2022-12-15
Use Ax plotting functions instead of Plotly in the Scheduler tutorial

esantorella:
Figures in the scheduler tutorial are not showing up because Plotly figures don't just automatically render properly with html. Instead we need to use the plotting functionality in Ax written for this.

Currently on the website on v:latest (stable is even more broken):
image

After this fix:
image

### DIFF
--- a/tutorials/scheduler.ipynb
+++ b/tutorials/scheduler.ipynb
@@ -421,7 +421,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


Balandat (commented):
> After this fix:

The image here looks suspiciously similar to the broken one - did you upload the right file?

esantorella:
> > After this fix:
> 
> The image here looks suspiciously similar to the broken one - did you upload the right file?

Oops, image is fixed.

Balandat (approved):
thanks!

PR #1328, opened 2022-12-15
Modify what runs on PRs that change tutorials

esantorella:
# What changed:
Old behavior on PRs that change tutorials:
* Those PRs run the nightly cron, in addition to the checks that run on all PRs
* The nightly cron runs tests, which is redundant since tests run on all PRs
* The nightly cron builds tutorials
* The nightly cron publishes the latest website. This has [never worked on PRs](https://github.com/facebook/Ax/actions/workflows/cron.yml?query=event%3Apull_request), even though it works in the nightly run, because the PRs are using a blank string in place of what should be a token from GH secrets. Presumably this is because [actions triggered by pull_request typically do not have access to secrets](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)
* The nightly cron tries to upload to test PyPi, which has never worked for the same reason

New behavior:
* PRs that run on tutorials build the tutorials, in addition to the checks that run on all PRs
* Behavior of the nightly cron is unchanged
* Tests are not run twice. No attempt to test uploading to PyPi or publish the website. Considering that this has never worked, I think we will be OK without that check.

The checks are running on this PR just for demo purposes because I made a modification to a tutorial, but that modification will be reverted.

image

# How
Added a new "tutorials" workflow that uses a "resusable tutorials" setup. The "reusable tutorials" setup is also used by the nightly cron. This is similar to what BoTorch does.

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -10,11 +10,6 @@ on:
     branches: [ main ]
     paths:
       - "tutorials/**"
-  pull_request:
-    branches: [ main ]
-    paths:
-      - "tutorials/**"
-
 
 jobs:
 
@@ -68,22 +63,8 @@ jobs:
         pytest -ra
 
   build-tutorials-with-pinned-botorch:
-
-    runs-on: ubuntu-latest
-
-    steps:
-    - uses: actions/checkout@v2
-    - name: Set up Python
-      uses: actions/setup-python@v2
-      with:
-        python-version: "3.8"
-    - name: Install dependencies
-      run: |
-        # will install the version of Botorch that is pinned in setup.py
-        pip install -e ".[tutorial]"
-    - name: Build tutorials
-      run: |
-        python scripts/make_tutorials.py -w $(pwd) -e
+    name: Build tutorials with pinned BoTorch
+    uses: ./.github/workflows/reusable_tutorials.yml
 
   publish-latest-website:
 
--- a/.github/workflows/reusable_tutorials.yml
+++ b/.github/workflows/reusable_tutorials.yml
@@ -1,0 +1,23 @@
+name: Reusable Tutorials Workflow
+
+on:
+  workflow_call:
+
+jobs:
+
+  build-tutorials-with-pinned-botorch:
+    name: Tutorials with pinned BoTorch
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: "3.8"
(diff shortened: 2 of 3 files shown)


tutorials/generation_strategy.ipynb
```diff
@@ -25,7 +25,7 @@
         "- Many higher-level APIs in Ax use generation strategies: Service and Loop APIs, `Scheduler` etc. (tutorials for all those higher-level APIs are here: https://ax.dev/tutorials/).\n",
         "- Generation strategy allows for storage and resumption of modeling setups, making optimization resumable from SQL or JSON snapshots.\n",
         "\n",
-        "This tutorial walks through a few examples of generation strategies and discusses its important settings. Before reading it, we recommend familiarizing yourself with how `Model` and `ModelBridge` work in Ax: https://ax.dev/docs/models.html#deeper-dive-organization-of-the-modeling-stack.\n",
+        "Thiss tutorial walks through a few examples of generation strategies and discusses its important settings. Before reading it, we recommend familiarizing yourself with how `Model` and `ModelBridge` work in Ax: https://ax.dev/docs/models.html#deeper-dive-organization-of-the-modeling-stack.\n",
```
Balandat, line 28:
```suggestion
        "This tutorial walks through a few examples of generation strategies and discusses its important settings. Before reading it, we recommend familiarizing yourself with how `Model` and `ModelBridge` work in Ax: https://ax.dev/docs/models.html#deeper-dive-organization-of-the-modeling-stack.\n",
```

.github/workflows/cron.yml
```diff
@@ -6,15 +6,6 @@ on:
...
   # allow this to be scheduled manually in addition to cron
   workflow_dispatch:
-  push:
-    branches: [ main ]
-    paths:
-      - "tutorials/**"
-  pull_request:
-    branches: [ main ]
-    paths:
-      - "tutorials/**"
```
Balandat, line 16: weird, was the whole cron job triggered by every PR before?

Balandat (approved):
gtmtl

.github/workflows/cron.yml
esantorella: Just the PRs that change tutorials

.github/workflows/cron.yml
```diff
@@ -6,15 +6,6 @@ on:
     - cron:  '0 5 * * *'
   # allow this to be scheduled manually in addition to cron
   workflow_dispatch:
```
esantorella, line 8:
```suggestion
  workflow_dispatch:
  push:
    branches: [ main ]
    paths:
      - "tutorials/**"
```

PR #1331, opened 2022-12-20
Avoid re-fitting SAASBO models when there's no new data

saitcakmak:
Summary: Sets `use_update=True` for SAASBO, which leads to skipping model fitting if there's no new data in `scheduler._fit_or_update_current_model`.

Reviewed By: lena-kashtelyan

Differential Revision: D42160763


saitcakmak:
Abandoned diff.

PR #1337, opened 2022-12-21
Clean up fast computations

saitcakmak:
Summary: https://github.com/pytorch/botorch/pull/1547 turned these off by default. No need to turn them off manually anymore.

Reviewed By: dme65, Balandat

Differential Revision: D42173976


saitcakmak:
this already landed

PR #1341, opened 2022-12-23
Update `BestPointMixin.get_trace` to extract observations directly from data

saitcakmak:
Summary: The reliance on a ModelBridge to convert data into tensors was leading to inconsistencies in the results due to the resulting tensors not being sorted by trial indices, which is needed for `get_trace`. This implements a lightweight helper for extracting the Y-tensor directly from the data that keeps the output sorted by the `trial_index`.

Reviewed By: Balandat

Differential Revision: D42221005


saitcakmak:
already committed

PR #1349, opened 2022-12-29
Website build: remove copy for missing files, make script fault tolerant

saitcakmak:
With exit on error removed: https://github.com/facebook/Ax/actions/runs/3802379960

With exit on error put back: https://github.com/facebook/Ax/actions/runs/3802399829

Still failing. Removed these from parse_sphinx as well: https://github.com/facebook/Ax/actions/runs/3802572734

With extra logging: https://github.com/facebook/Ax/actions/runs/3802664350

### DIFF
--- a/scripts/make_docs.sh
+++ b/scripts/make_docs.sh
@@ -4,6 +4,9 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
+# Exit if any error occurs
+set -e
+
 usage() {
   echo "Usage: $0 [-b] [-o] [-r] [-t] [-k kernel_name]"
   echo ""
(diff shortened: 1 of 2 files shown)


scripts/make_docs.sh
```diff
@@ -4,9 +4,6 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
-# Exit if any error occurs
-set -e
-
```
saitcakmak, line 9: This might not be ideal since we run the tutorials against latest as part of this script -- which I am not a fan of --

saitcakmak:
These tests I ran do not use the updated script since `publish_site.sh` pulls the main branch and uses the script from there. So, we need to land the diff to test the changes. Let's re-visit after the release.

PR #1350, opened 2022-12-29
Use stable botorch in deploy workflow

saitcakmak:
We can be in situations where we're releasing version of Ax that is compatible with pinned BoTorch but is not fully compatible with latest BoTorch. This should eliminate the CI failing in such cases


saitcakmak:
Abandoned

PR #1366, opened 2023-01-06
Always clear "data/MNIST/raw" in test tearDowns

danielcohenlive:
Summary: this is the fully automatic solution.  It only takes `0.0003138110041618347s`...

Reviewed By: mpolson64

Differential Revision: D42387374


saitcakmak:
abandoned

PR #1380, opened 2023-01-18
Fix test_extract_Y_from_data

saitcakmak:
Summary: Since the first two trials were not completed, when the missing data corresponded to those, the error would not be raised.

Differential Revision: D42564886


Balandat (approved):
thanks

PR #1383, opened 2023-01-19
Fix flaky test_GetPosteriorMean

danielcohenlive:
Summary: It's timing out on gen, so I'm using `fast_botorch_optimize` and it's working like magic!

Reviewed By: pcanaran

Differential Revision: D42608149


Balandat:
This was merged in as https://github.com/facebook/Ax/commit/3132a55d9e154270d2b410b02d8da25fbf7bd49c

PR #1399, opened 2023-01-30
Skip Cast in TorchModelBridge._untransform_objective_thresholds

saitcakmak:
Summary:
Cast requires a complete heirarchical parameterization and will error out otherwise. This attempts to transform an incomplete parameterization.

It looks like `fixed_features_obs` is only needed for the `StratifiedStandardizeY` transform, so filtering based on that would be another option.

NOTE: This is the only use of  `untransform_observation_features` with incomplete parameterizations.

Reviewed By: lena-kashtelyan

Differential Revision: D42810742


saitcakmak:
already merged

PR #1414, opened 2023-02-10
Fixed bug where optimize_acqf didn't work with different batch sizes

esantorella:
Differential Revision: D43178298


saitcakmak:
merged

PR #1418, opened 2023-02-14
Feature/json decoder with pathlib path and functions

madeline-scyphers:
# Summary

## Problem
The current `CORE_DECODER_REGISTRY` and `CORE_ENCODER_REGISTRY` does not include `pathlib.Path` instances which people often store paths as. I am trying to encode and decode some custom Ax Metrics and other subclasses that include me sometime storing paths. A problem with the current encoding and decoding options is that they need to be classes (instead of a function) because of this line in `object_from_json`
``` python
elif issubclass(_class, SerializationMixin):
```
Which doesn't work with `pathlib.Path` since it also only takes position arguments that get splatted/unpacked. It would be easy to to make it accept functions functions by making the check `elif isclass(_class) and issubclass(_class, SerializationMixin):` which happens for other similar checks a dozen or so lines above this line.

Doing so would also allow encoders and decoders that are `classmethods` which are often used as alternative constructors (Something.from_json, Something.to_json, etc.) 

## Changes

- Changed `object_from_json` to allow functions  to be in the `decoder_registry`. 
  - This was done as described above by changing `elif issubclass(_class, SerializationMixin):` to `elif isclass(_class) and issubclass(_class, SerializationMixin):`
- add pathlib.Path to `CORE_DECODER_REGISTRY` and `CORE_ENCODER_REGISTRY` 
-  Adjust typehints around decoder registries from `Dict[str, Type]` to `Dict[str, Callable[[Dict[str, Any]], Any]]`
- Add `pathlib.Path` to `TEST_CASES` in test_json_store.py


mpolson64:
Thank you so much for adding this! I've imported it into our infra and it passes all* our internal tests -- I'm happy to accept this and merge it in.

The one hitch is it looks like our type checking is failing. Feel free to put pyre fixmes for any errors relating to Any appearing in any parametric types. Let me know if you have any questions, and once you're able to get `pyre` passing lets get this landed :)

madeline-scyphers:
@mpolson64 I tried getting Pyre set up, but when I do, it errors with hundreds of type hint problems, which I assume isn't right. So I fixed what you mentioned and I think I fixed it everywhere I needed to, but I can't verify since I couldn't get pyre to work, and pyre isn't ran as part of your public facing CI jobs.

mpolson64:
@madeline-scyphers My apologies! There must be some discrepancies in our our Pyre configurations (on our internal development environment Pyre passes with no type checking errors). I'll go ahead and do the following:
- Import your PR to our internal setup and see if I can make our Pyre happy
- Set up a working Pyre config in our github so this doesn't happen in the future
I'll try and do the former within the next week or so but the latter may have to wait a while -- thanks for bearing with us through this whole process.

PR #1431, opened 2023-02-19
fixing the derivative gp posterior for matern kernel

belakaria:
The previous implementation had a distinction between Matern with l1 distance and l2 distance. This distinction is not accurate and is unnecessary.


saitcakmak:
Hi Syrine, thanks for the fix. If you have any references for the derivative formulas, it'd be great to add them to the code. I'll ask @bletham to review the changes internally.

belakaria:
Hi @saitcakmak. I don't have a reference for this specific derivation. I did it myself. But I will share the derivation with Ben.

Issue #1439, opened 2023-02-21
Use of seed for `get_sobol` not producing identical sets when repeated


sgbaird:
modifying the snippet from https://github.com/facebook/Ax/issues/1437
```python
SEED = 10
parameters = [
    {"name": "mu1", "type": "range", "bounds": [1.0, 5.0], "value_type": "float"},
    {"name": "mu2", "type": "range", "bounds": [1.0, 5.0], "value_type": "float"},
    {"name": "mu3", "type": "range", "bounds": [1.0, 5.0], "value_type": "float"},
    {"name": "std1", "type": "range", "bounds": [0.1, 1.0], "value_type": "float"},
    {"name": "std2", "type": "range", "bounds": [0.1, 1.0], "value_type": "float"},
    {"name": "std3", "type": "range", "bounds": [0.1, 1.0], "value_type": "float"},
    {"name": "comp1", "type": "range", "bounds": [0.0, 1.0], "value_type": "float"},
    {"name": "comp2", "type": "range", "bounds": [0.0, 1.0], "value_type": "float"},
    {"name": "comp3", "type": "range", "bounds": [0.0, 1.0], "value_type": "float"},
    {"name": "num_particles", "type": "range", "bounds": [100, 1000]},
    {"name": "safety_factor", "type": "range", "bounds": [1.0, 2.5]},
]
ax_client = AxClient()
ax_client.create_experiment(
    name="boppf_sobol",
    parameters=parameters,
    objective_name="packing_fraction",
    minimize=False,
    parameter_constraints=["std1 <= std2", "std2 <= std3"],
)
search_space = ax_client.experiment.search_space
m = get_sobol(search_space, fallback_to_sample_polytope=True, seed=SEED)
gr = m.gen(n=num_samples)
param_df = pd.DataFrame([arm.parameters for arm in gr.arms])

print(param_df.shape[0])
# 65536

m2 = get_sobol(search_space, fallback_to_sample_polytope=True, seed=SEED)
gr2 = m2.gen(n=num_samples)
param_df2 = pd.DataFrame([arm.parameters for arm in gr2.arms])
```

After merging the two repeat datasets, there are no duplicates.
```python
print(pd.concat((param_df, param_df2)).drop_duplicates().shape[0])
# 131072
```

Even with aggressive rounding, only a couple hundred repeats out of over 100k entries are present.
```python
print(pd.concat((param_df2.round(3), param_df3.round(3))).drop_duplicates().shape[0])
# 130900
```

[...]

PR #1441, opened 2023-02-21
add clarification about polytope sampler

sgbaird:
i.e., that polytope sampling and Sobol sampling are different algorithms https://github.com/facebook/Ax/issues/1439#issuecomment-1439195984

### DIFF
--- a/ax/models/random/base.py
+++ b/ax/models/random/base.py
@@ -135,7 +135,7 @@ class RandomModel(Model):
             if self.fallback_to_sample_polytope:
                 logger.info(
                     "Rejection sampling exceeded specified maximum draws."
-                    "Falling back on polytope sampler"
+                    "Falling back on polytope sampler instead of Sobol sampler"
                 )
                 # If rejection sampling fails, try polytope sampler.
                 polytope_sampler = HitAndRunPolytopeSampler(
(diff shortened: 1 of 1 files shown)


Balandat (approved):
thanks

ax/models/random/base.py
```diff
@@ -135,7 +135,7 @@ def gen(
             if self.fallback_to_sample_polytope:
                 logger.info(
                     "Rejection sampling exceeded specified maximum draws."
-                    "Falling back on polytope sampler"
+                    "Falling back on polytope sampler instead of Sobol sampler"
```
saitcakmak, line 138:
```suggestion
                    "Falling back on polytope sampler instead of Sobol sampler."
```

ax/models/random/base.py
```diff
@@ -135,7 +135,7 @@ def gen(
             if self.fallback_to_sample_polytope:
                 logger.info(
                     "Rejection sampling exceeded specified maximum draws."
-                    "Falling back on polytope sampler"
+                    "Falling back on polytope sampler instead of Sobol sampler."
```
saitcakmak, line 138: Looks like this is on the base class, which is shared between uniform & sobol samplers. 
```suggestion
                    "Falling back on HitAndRunPolytopeSampler instead of "
                    f"{self.__class__.__name__}."
```

PR #1487, opened 2023-03-04
Fix Acquisition.get_botorch_objective_and_transform

saitcakmak:
Summary: This does not pass down the acqf class to the `get_botorch_objective_and_transform` helper, and always passes down `objective_thresholds` if they're available. As a result, `get_botorch_objective_and_transform` determines the type of the objective to construct (MOO or otherwise) based on the availability of `objective_thresholds`. This leads to weird downstream errors if objective thresholds are present when a single objective acquisition function is being used. This change should fix that.

Differential Revision: D43803349


saitcakmak:
abandoned (alternative fix landed)

PR #1497, opened 2023-03-08
Handle the case of more unordered choices than ordered parameters in choose_gs

saitcakmak:
Summary:
Previously, this would just dispatch to SOBOL, which is not ideal if the choice parameters have many options. We should be either using standard BO or BO MIXED.

This implementation favors standard BO options, though I don't have a good reason why this shouldn't be using BO MIXED.

Reviewed By: dme65

Differential Revision: D43724682


saitcakmak:
Already merged

PR #1510, opened 2023-03-15
Remove checked_cast_complex

mpolson64:
Summary: typeguard updated to 3.0.0 which broke backwards incompatiblity and caused a huge headache for our OSS users. Rather than pin OSS to a stale version (or god forbid try and update typeguard for all fbcode!) let's just remove this functionality. Follow up diff will remove typeguard from kwargs.py

Reviewed By: Balandat

Differential Revision: D44095802


jakeBass:
Hi folks! It appears that Ax still has [pinned the typeguard requirement to an ancient version](https://github.com/facebook/Ax/blob/main/setup.py#L31). Are there any plans to revive this PR, or to otherwise update the typeguard requirement?

PR #1537, opened 2023-03-21
Filter out data standardization warnings in cross_validate

saitcakmak:
Summary:
We standardize the full training observations, then use only a subset of these in each CV fold. Since we removed some points, the remaining observations have a different mean and std dev, which leads to them failing the standardization test and raising `InputDataWarning: Input data is not standardized. Please consider scaling the input to zero mean and unit variance.`.

Filtering out this warning in CV will help avoid confusion and clear up some unnecessary logs.

Reviewed By: dme65

Differential Revision: D44262090

### DIFF
--- a/ax/modelbridge/base.py
+++ b/ax/modelbridge/base.py
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 import time
+import warnings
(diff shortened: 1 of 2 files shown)


saitcakmak:
committed

PR #1554, opened 2023-03-27
Add helper function to get the fit_time and gen_time in each trial (not just aggregate)

esantorella:
Summary: This will be useful for benchmarking

Differential Revision: D44434115

### DIFF
--- a/ax/core/tests/test_utils.py
+++ b/ax/core/tests/test_utils.py
@@ -19,10 +19,14 @@ from ax.core.utils import (
     best_feasible_objective,
     get_missing_metrics,
     get_missing_metrics_by_name,
+    get_model_times,
+    get_model_trace_of_times,
     MissingMetrics,
 )
 from ax.modelbridge.modelbridge_utils import feasible_hypervolume
 from ax.utils.common.testutils import TestCase
+from ax.utils.testing.core_stubs import get_robust_branin_experiment
+from pyre_extensions import none_throws
 
 
 class UtilsTest(TestCase):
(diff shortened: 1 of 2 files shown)


ax/core/tests/test_utils.py
```diff
@@ -185,3 +189,14 @@ def test_feasible_hypervolume(self) -> None:
...
         self.assertEqual(list(feas_hv), [0.0, 0.0, 1.0, 1.0])
+
+    def test_get_model_times(self) -> None:
+        exp = get_robust_branin_experiment(num_sobol_trials=2)
+        fit_times, gen_times = get_model_trace_of_times(exp)
+        total_fit_time, total_gen_time = get_model_times(exp)
+        fit_times_not_none = [none_throws(elt) for elt in fit_times]
+        gen_times_not_none = [none_throws(elt) for elt in gen_times]
+        self.assertTrue([elt > 0 for elt in fit_times_not_none])
+        self.assertTrue([elt > 0 for elt in gen_times_not_none])
```
Balandat, line 200: This only asserts that the list is non-empty (i.e. this passes even if every list element is `False`) - is this what you want? Or do you want to assert that `all(elt > 0 for elt in fit_times_not_none)` is `True`?

ax/core/utils.py
```diff
@@ -148,21 +150,37 @@ def best_feasible_objective(  # pragma: no cover
     return accumulate(f)
 
 
+def _extract_generator_run(trial: BaseTrial) -> GeneratorRun:
+    if isinstance(trial, BatchTrial):
+        return trial._generator_run_structs[0].generator_run
```
Balandat, line 155: This assumes the trial was generated by one generator run (in some cases there may be more, like in the online experimentation setting). We should probably raise an informative `UnsupportedError` or `NotImplementedError` here if that's not the case (and we don't want to implement the general case for now).

ax/core/utils.py
```diff
@@ -148,21 +150,37 @@ def best_feasible_objective(  # pragma: no cover
...
-            gr = not_none(trial.generator_run)
-        else:
-            raise ValueError("Unexpected trial type")  # pragma: no cover
-        if gr.fit_time is not None:
-            fit_time += not_none(gr.fit_time)
-        if gr.gen_time is not None:
-            gen_time += not_none(gr.gen_time)
+    fit_times, gen_times = get_model_trace_of_times(experiment)
+    fit_time = sum([t for t in fit_times if t is not None])
+    gen_time = sum([t for t in gen_times if t is not None])
```
Balandat, line 185: Can sum directly on a generator
```suggestion
    fit_time = sum(t for t in fit_times if t is not None)
    gen_time = sum(t for t in gen_times if t is not None)
```

ax/core/tests/test_utils.py
esantorella: Oops, yes, I wanted the latter!

PR #1576, opened 2023-04-09
Efficiency improvement of generate_all_input_matrix in Sobol sensitivity analysis

SebastianAment:
Summary: `generate_all_input_matrix` currently iteratively concatenates tensors in a loop. It is more efficient to first create the component tensors individually, and then concatenate them once to create a single large tensor at the end of the function. This is the purpose of this commit.

Differential Revision: D44817702


Balandat (approved):
Lgtm

PR #1591, opened 2023-04-24
Set logging level to WARNING for unit tests

esantorella:
The default Ax logging level is INFO, and Ax prints a lot of INFO logs. This is annoying for unit tests, where we only want to know when something is wrong, and drowns out useful warnings. This sets the logging level one level higher, to WARNING and above.

# Test plan

Ran unit tests and made sure that INFO logs were not printed, but WARNING logs were.

### DIFF
--- a/ax/utils/common/testutils.py
+++ b/ax/utils/common/testutils.py
@@ -12,6 +12,7 @@
 import contextlib
 import io
 import linecache
+import logging
 import signal
 import sys
 import types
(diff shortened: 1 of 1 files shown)


ax/utils/common/testutils.py
```diff
@@ -299,6 +300,15 @@ def signal_handler(signum: int, frame: Optional[FrameType]) -> None:
         super().__init__(methodName=methodName)
         signal.signal(signal.SIGALRM, signal_handler)
 
+    def setUp(self) -> None:
+        """
+        Only show log messages of WARNING or higher while testing.
+
+        Ax prints a lot of INFO logs that are not relevant for unit tests.
+        """
+        logger = get_logger(__name__, level=logging.WARNING)
```
esantorella, line 309:
```suggestion
        # Parent handlers are shared, so setting the level this
        # way applies it to all Ax loggers.
        logger = get_logger(__name__, level=logging.WARNING)
```

PR #1592, opened 2023-04-25
remove libfb import from test_strategies test

rsshah1993:
Summary: open source doesn't have libfb dependency so removing from tests

Differential Revision: D45271241


saitcakmak:
abandoned

PR #1610, opened 2023-05-08
Add Global Stopping Strategy Tutorial

esantorella:
Summary: This only uses the Service API. GSS is also supported by Scheduler, but I didn't include this since a minimal example with `Scheduler` is pretty involved. I think it would make sense to add a little bit about GSS into the existing scheduler tutorial, to avoid a lot of unnecessary duplication.

Differential Revision: D45671161


saitcakmak (approved):
LGTM!

PR #1615, opened 2023-05-10
Fix DeprecationWarning exception in Raytune tutorial

esantorella:
The `raytune_pytorch_cnn_tutorial` has been failing due to a `DeprecationWarning` that is raised as an exception. This PR
* Updates import paths so that there is no DeprecationWarning
* Updates the readme to recommend that developers install the "tutorials" dependencies, matching the text advice to install "all" dependencies
* Adds the option to run just one tutorial by name to the `make_tutorials.py` script, matching the BoTorch tutorials script.

# Test plan:

CI should pass now.

### DIFF
--- a/scripts/make_tutorials.py
+++ b/scripts/make_tutorials.py
@@ -61,7 +61,10 @@ function require(deps, fxn) {
 
 
 def gen_tutorials(
-    repo_dir: str, exec_tutorials: bool, kernel_name: Optional[str] = None
+    repo_dir: str,
+    exec_tutorials: bool,
+    kernel_name: Optional[str] = None,
+    name: Optional[str] = None,
 ) -> None:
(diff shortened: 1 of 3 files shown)


scripts/make_tutorials.py
```diff
@@ -61,7 +61,10 @@ class TutorialPage extends React.Component {{
 
 
 def gen_tutorials(
-    repo_dir: str, exec_tutorials: bool, kernel_name: Optional[str] = None
+    repo_dir: str,
+    exec_tutorials: bool,
+    kernel_name: Optional[str] = None,
+    name: Optional[str] = None,
```
Balandat, line 67: Could potentially also have `names: Optional[List[str]]` to select more than one tutorial if that would be useful.

Balandat (approved):
thanks for fixing this!

PR #1646, opened 2023-06-05
Implement ArtifactData

bernardbeckerman:
Summary: Adds `ArtifactData` which is a child class of `BaseData` that has a required "value" field in place of `Data`'s required "mean" and "sem" fields.

Differential Revision: D46239752


saitcakmak:
abandoned

PR #1650, opened 2023-06-08
Avoid spurious UserWarning in get_pareto_frontier_and_configs

Balandat:
Prior to this change, using `AxClient` for multi-objective optimization would needlessly spam the logs with a `UserWarning` about deprecated use of `transform_outcomes_and_configs` in `get_pareto_frontier_and_configs`. The reason is that #1200 did not update all the call sites.


esantorella (approved):

esantorella:
Thanks!

PR #1651, opened 2023-06-08
Issue 1628 constrained get optimization trace

### LINKED ISSUE #1628: ax_client.get_optimization_trace() does not discard unfeasible trials
Hello! I'm using the service API to solve a single-objective optimization with an outcome constraint defined by `outcome_constraint=["SLA >= 0.9"]`.

So far so good. Upon inspection, I see that the `ax_client.get_trials_data_frame()` correctly identifies feasible/unfeasible trials as desired. The issue is when I use the `render(ax_client.get_optimization_trace())` function. The plot it outputs does consider unfeasible points and also does not offer an option to discard them.

Is that a design choice? If not, can we add a parameter to do it?

I'm not familiar with the code base yet but I can give it a shot if you guys think it is useful.

I'm using `version 0.3.2`.

leandrobbraga:
This MR solves #1628 which did not consider the infeasible results while calculating the best values so far.

I'm very new to the code base and  would like help in two points:
1) What's the best way to check for feasiblility, is this way correct?
2) Is the way I checked the plot the best way?

Closes #1628

### DIFF
--- a/ax/service/ax_client.py
+++ b/ax/service/ax_client.py
@@ -56,6 +56,7 @@ from ax.exceptions.core import (
 )
 from ax.exceptions.generation_strategy import MaxParallelismReachedException
 from ax.global_stopping.strategies.base import BaseGlobalStoppingStrategy
+from ax.global_stopping.strategies.improvement import constraint_satisfaction
 from ax.modelbridge.dispatch_utils import choose_generation_strategy
 from ax.modelbridge.generation_strategy import GenerationStrategy
 from ax.modelbridge.modelbridge_utils import (
@@ -924,11 +925,19 @@ class AxClient(WithDBSettingsBase, BestPointMixin, InstantiationBase):
                 "for multi-objective experiments"
             )
 
+        # We wan't to show the objective value for a trial only if it satisfies
+        # the constraints
+        def _constrained_trial_objective_mean(trial: Trial) -> float:
+            if constraint_satisfaction(trial):
+                return checked_cast(Trial, trial).objective_mean
+            else:
+                return np.inf if objective.minimize else -np.inf
+
         objective_name = self.objective_name
         best_objectives = np.array(
             [
                 [
-                    checked_cast(Trial, trial).objective_mean
+                    _constrained_trial_objective_mean(trial)
                     for trial in self.experiment.trials.values()
                     if trial.status.is_completed
                 ]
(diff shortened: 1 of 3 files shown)


ax/service/tests/test_ax_client.py
```diff
@@ -2562,61 +2562,6 @@ def get_solution(use_y0_threshold: bool, use_y2_constraint: bool) -> Set[int]:
                 return {0, 1}
             return {0, 2}
 
-            ax_client = get_client_with_simple_discrete_moo_problem(
```
leandrobbraga, line 2565: If you wonder why I removed this, this is unreachable, is this meant this way?

esantorella: Wow, good catch! I'm pretty sure this code is supposed to be run. It was accidentally indented too far during a refactoring. Could you unindent it?

esantorella (changes requested):
Thank you for this! This seems like the correct design to me, and a beneficial change. I made a couple small suggestions but once those are fixed this should be good to go.

ax/service/tests/test_ax_client.py
```diff
@@ -1441,13 +1441,13 @@ def test_raw_data_format_with_map_results(self) -> None:
             )
 
     def test_keep_generating_without_data(self) -> None:
-        # Check that normally numebr of arms to generate is enforced.
+        # Check that normally number of arms to generate is enforced.
```
esantorella, line 1444: Thank you for fixing the typos!

ax/service/ax_client.py
```diff
@@ -924,11 +925,19 @@ def get_optimization_trace(
                 "for multi-objective experiments"
             )
 
+        # We wan't to show the objective value for a trial only if it satisfies
+        # the constraints
+        def _constrained_trial_objective_mean(trial: Trial) -> float:
+            if constraint_satisfaction(trial):
+                return checked_cast(Trial, trial).objective_mean
+            else:
+                return np.inf if objective.minimize else -np.inf
```
esantorella, line 934:
```suggestion
        # Setting the objective values of infeasible points to be infinitely
        # bad prevents them from increasing or decreasing the
        # optimization trace.
        def _constrained_trial_objective_mean(trial: Trial) -> float:
            if constraint_satisfaction(trial):
                return checked_cast(Trial, trial).objective_mean
            return np.inf if objective.minimize else -np.inf
```

PR #1666, opened 2023-06-20
retitle to "Parameter Importance"

bernardbeckerman:
Summary:
Old terminology "feature importance" is confusing and somewhat overloaded with the standard ML practitioner's concept of feature importance. This should make these plots more usable.

NOTE: this diff is specifically not performing a more sweeping conversion of "feature importance" -> "parameter importance" more generally throughout the codebase, which would involve changes up and down the stack, and is overkill for the current purpose of making the plots more legible by downstream users. Please comment on this diff if you think this sweeping change should follow.

Reviewed By: Balandat

Differential Revision: D46801256


saitcakmak:
committed

PR #1667, opened 2023-06-20
change "Total Sobol indices" -> "Variance-based sensitivity analysis"

bernardbeckerman:
Summary: Users were getting confused with the term "Sobol indices" since Sobol did so many cool things (particularly Sobol sequences, which are used to generate exploratory trials at the beginning of a sweep and are referred to by name, thus the confusion). Switching to referring to this technique using the name of the corresponding [wikipedia article title](https://en.wikipedia.org/wiki/Variance-based_sensitivity_analysis) (sorry Ilya!).

Reviewed By: Balandat

Differential Revision: D46801454


saitcakmak:
already committed

PR #1708, opened 2023-07-10
Allow passing jit_compile as part of mll_options

saitcakmak:
Summary:
Currently, we cannot specify `jit_compile` in MBM when fitting fully-Bayesian models. Since this is an argument exposed in `choose_generation_strategy`, we need to support it :)

This uses `mll_options` to pass in the arg, which avoids adding a new field to `SurrogateSpec`. Another option is to add a `model_fit_kwargs` field to `SurrogateSpec` and pass this down to `fit_botorch_model` and use it for both fully-Bayesian and other models from there. Happy to implement that if that's preferred.

Reviewed By: dme65

Differential Revision: D47349459


saitcakmak:
committed

PR #1748, opened 2023-07-28
Update deprecated Pandas and Numpy functionality

esantorella:
* The "interpolation" argument to `np.percentile` and `np.nanpercentile` is now "method."
* Pandas will soon not allow using `float` on Series; if there's a single element in the series, you now need to explicitly extract it first. I fixed this in `extract_Y_from_data` and also made the logic and error messages clearer.

# Test plan:

Units.

### DIFF
--- a/ax/service/tests/test_best_point_utils.py
+++ b/ax/service/tests/test_best_point_utils.py
@@ -408,13 +408,27 @@ class TestBestPointUtils(TestCase):
 
         # Error with missing data.
         with self.assertRaisesRegex(
-            UserInputError, "single data point for each metric"
+            UserInputError, "Trial 2 is missing data on metrics {'foo'}."
         ):
             # Skipping first 5 data points since first two trials are not completed.
+
             extract_Y_from_data(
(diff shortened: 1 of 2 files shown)


ax/modelbridge/transforms/winsorize.py
```diff
@@ -422,14 +422,14 @@ def _quantiles_to_cutoffs(
     elif lower == 0.0:  # Use the default cutoff if there is no winsorization
         cutoff_l = DEFAULT_CUTOFFS[0]
     else:
-        cutoff_l = np.percentile(Y, lower * 100, interpolation="lower")
+        cutoff_l = np.percentile(Y, lower * 100, mthod="lower")
```
saitcakmak, line 425:
```suggestion
        cutoff_l = np.percentile(Y, lower * 100, method="lower")
```

saitcakmak (approved):
Thanks! I guess this means we require numpy >1.22 going forward. 1.22 was released 1.5 years ago, so I think that's totally fine.

PR #1750, opened 2023-07-30
[WIP] Require exact equality for Pandas DataFrames to be considered equal

esantorella:
While debugging a numerical issue, I encountered a case where an experiment had float64 trial data where the data changed values after being saved and reloaded. It looks like the data was converted to float32 and then back to float64. However, equality checks still passed because Ax is only testing that values are approximately equal, to 1e-8.

I changed tests so that this failure is surfaced, but still need to fix it.

# Test plan

Changed a unit test so that this failure is triggered.

### DIFF
--- a/ax/storage/json_store/tests/test_json_store.py
+++ b/ax/storage/json_store/tests/test_json_store.py
@@ -364,7 +364,7 @@ class JSONStoreTest(TestCase):
                 class_encoder_registry=CORE_CLASS_ENCODER_REGISTRY,
             )
 
-        # Key error on desearialization.
+        # Key error on deserialization.
         x_json = object_to_json(
             x,
             encoder_registry=CORE_ENCODER_REGISTRY,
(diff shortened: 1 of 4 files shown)


Balandat:
> I encountered a case where an experiment had float64 trial data where the data changed values after being saved and reloaded. It looks like the data was converted to float32 and then back to float64

Is this something that happened on the Ax side? I've run into floating point precision issues in the past (https://github.com/facebook/Ax/pull/1538), ideally we could fix this at the root of the issue.

esantorella:
> Is this something that happened on the Ax side? I've run into floating point precision issues in the past (#1538), ideally we could fix this at the root of the issue.

Yeah, this had to do with pandas `to_json`. I'm not sure if this would have avoided the issue behind #1538  -- depends whether the numbers involved came from serializing a DataFrame.

PR #1757, opened 2023-08-02
Allow "unsafe" trial completion in AxClient

saitcakmak:
Summary:
This adds an option to unsafely complete a trial in AxClient. This will
- Ignore & overwrite any previously attached data for the given trial
- Mark trial as completed regardless of its previous status (default behavior expects the trial to be running)
- Avoid errors that are currently raised if we try to complete an already completed trial
- Allow completing a trial that was previously failed or abandoned

Differential Revision: D48009332


saitcakmak:
abandoned

PR #1762, opened 2023-08-04
Update tutorial dependencies

saitcakmak:
PyArrow is required to avoid the tutorial failures seen in the nightly cron: https://github.com/facebook/Ax/actions/runs/5758595848/job/15611390170

Smoke-test tutorial runs:
On current main: https://github.com/facebook/Ax/actions/runs/5764907837
On this PR: https://github.com/facebook/Ax/actions/runs/5764906547


mpolson64:
Thanks for catching this! I was just about to put a diff up myself but this is perfect.

PR #1776, opened 2023-08-10
update installation instructions

sdaulton:
Summary: follow up on comment from D48230718

Differential Revision: D48245744


saitcakmak:
committed

PR #1777, opened 2023-08-10
bump pinned botorch version to 0.9.2

sdaulton:
Summary: Bump pinned version to include hotfix for constraints in MBM with single objective acquisition function

Reviewed By: saitcakmak

Differential Revision: D48230718


saitcakmak:
committed

PR #1789, opened 2023-08-18
fix bug with integers involved in parameter constraints

sdaulton:
Summary:
see title.

This handles the edge case in IntToFloat, where parameter constraints are placed on integers, but not all integers are present in `untransform_observation_features`.

Differential Revision: D48482657


Balandat (approved):
lgtm, thanks!

PR #1799, opened 2023-08-25
[Unit test fix] Fix BoTorch MOO mock imports 

esantorella:
Tests are [failing](https://github.com/facebook/Ax/actions/runs/5972071187/job/16202069980) with pinned BoTorch as a result of BoTorch #1986 / Ax #1783, because import paths were updated.

# Test plan:

Existing units; add a task to revert this once BoTorch 0.9.3 is pinned.


esantorella:
I tested that this works with the pinned BoTorch version on my laptop. The CI tests on BoTorch "latest" on PR.

PR #1802, opened 2023-08-28
Store bounds in double precision in test functions

esantorella:
Summary:
## Motivation

* BoTorch recommends double precision in optimization, and will warn on single-precision inputs.
* `torch.tensor(a_float)` will use single precision by default. This is how bounds are currently instantiated in test problems.
* Native floats are already double precision, so this is actually downcasting them.

X-link: https://github.com/pytorch/botorch/pull/1996

Reviewed By: saitcakmak

Differential Revision: D48757569

Pulled By: esantorella


saitcakmak:
committed

PR #1810, opened 2023-08-31
remove errant ftest (focused test for tsloop)

bernardbeckerman:
Summary:
These get left in from time to time. Not sure how D48829031 landed given that the linter usually stops this.

Created from CodeHub with https://fburl.com/edit-in-codehub

Reviewed By: mpolson64

Differential Revision: D48883557


saitcakmak:
already merged

PR #1822, opened 2023-09-06
Update GitHub Actions Workflows

saitcakmak:
- Update `checkout` and `python-setup` to v3 & v4, respectively.
- Create a new `reusable_test` workflow and use this everywhere unit tests are ran. This supports pinned and latest BoTorch, as well as minimal and full dependencies.
- Update the `reusable_tutorials` workflow to support both latest and pinned BoTorch.
- Create a `cron_pinned` workflow and update `cron` to run only latest BoTorch version. This should make nightly cron into a much higher signal workflow and prevent it from failing due to latest Ax diverging from pinned BoTorch.

### DIFF
--- a/.github/workflows/reusable_test.yml
+++ b/.github/workflows/reusable_test.yml
@@ -38,7 +38,7 @@ jobs:
     - if: ${{ inputs.pinned_botorch }}
       name: Install dependencies with pinned BoTorch (minimal dependencies ${{ inputs.minimal_dependencies }})
       run: |
-        pip install -e ${{ (inputs.minimal_dependencies) && ".[unittest_minimal]") || ".[unittest]" }}
+        pip install -e ${{ (inputs.minimal_dependencies) && '.[unittest_minimal]') || '.[unittest]' }}
 
     - if: ${{ !inputs.pinned_botorch }}
       name: Install dependencies with latest BoTorch (minimal dependencies ${{ inputs.minimal_dependencies }})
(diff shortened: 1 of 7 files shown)


.github/workflows/reusable_test.yml
```diff
@@ -0,0 +1,65 @@
...
+        required: false
+        type: boolean
+        default: false
+
+jobs:
+  tests-and-coverage:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.9", "3.10"]
```
Balandat, line 28: Should we start testing against 3.11 (here and in other places)? Botorch does...

.github/workflows/reusable_test.yml
```diff
@@ -0,0 +1,65 @@
...
+    - uses: actions/checkout@v3
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - if: ${{ inputs.pinned_botorch }}
+      name: Install dependencies with pinned BoTorch (minimal dependencies ${{ inputs.minimal_dependencies }})
+      run: |
+        pip install -e ${{ ((inputs.minimal_dependencies) && '.[unittest_minimal]') || '.[unittest]' }}
```
Balandat, line 41: can you add a comment for what's going on here?

Balandat (approved):
lgtm, thanks!

.github/workflows/reusable_test.yml
saitcakmak: We should but this is currently blocked on https://github.com/pytorch/torchx/issues/744

.github/workflows/reusable_test.yml
saitcakmak, line 41:
```suggestion
        # The brackets returns '.[unittest_minimal]' if using minimal dependencies and '.[unittest]'
        # otherwise. This saves us from needing 4 install dependencies blocks by supporting two
        # different installation options in one line.
        pip install -e ${{ ((inputs.minimal_dependencies) && '.[unittest_minimal]') || '.[unittest]' }}
```

PR #1826, opened 2023-09-08
Fix smoke test for SEBO tutorial

saitcakmak:
For the AxClient parts, this was returning noise levels but using `SingleTaskGP`, which doesn't accept `train_Yvar`. Leaving the model type as `None` should dispatch the correct model type.

Test Plan: https://github.com/facebook/Ax/actions/runs/6117752845


Balandat (approved):
makes sense

PR #1827, opened 2023-09-08
Fix naming of reusable tutorials workflow

saitcakmak:
This was calling itself "Tutorials with pinned BoTorch" despite supporting both pinned & latest BoTorch. Updating the name to clear up the confusion.

Confusing naming before this: https://github.com/facebook/Ax/actions/runs/6117752845
No longer calling itself pinned while running latest: https://github.com/facebook/Ax/actions/runs/6117768518


Balandat (approved):
thanks!

PR #1848, opened 2023-09-14
Use mini batches in SobolSensitivityGPMean

saitcakmak:
Summary: In the existing setup, the cost of `SobolSensitivityGPMean` scales super-linearly in `num_mc_samples`. This change reduces the size of tensors created significantly, which should limit the peak memory usage.

Differential Revision: D49298572

### DIFF
--- a/ax/utils/sensitivity/sobol_measures.py
+++ b/ax/utils/sensitivity/sobol_measures.py
@@ -397,6 +397,7 @@ class SobolSensitivityGPMean(object):
         link_function: Callable[
             [torch.Tensor, torch.Tensor], torch.Tensor
         ] = GaussianLinkMean,
+        mini_batch_size: int = 128,
     ) -> None:
         r"""Computes three types of Sobol indices:
         first order indices, total indices and second order indices (if specified ).
@@ -411,6 +412,8 @@ class SobolSensitivityGPMean(object):
             input_qmc: If True, a qmc Sobol grid is use instead of uniformly random.
             num_bootstrap_samples: If bootstrap is true, the number of bootstraps has
                 to be specified.
+            mini_batch_size: The size of the mini-batches used while evaluating the
+                model posterior. Increasing this will increase the memory usage.
         """
         self.model = model
         self.dim: int = _get_input_dimensionality(model)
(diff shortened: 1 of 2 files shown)


ax/utils/sensitivity/tests/test_sensitivity.py
```diff
@@ -109,7 +109,7 @@ def testDgsmGpSampling(self) -> None:
         self.assertEqual(gradients_absolute_measure.shape, torch.Size([2, 5]))
         self.assertEqual(gradients_square_measure.shape, torch.Size([2, 5]))
 
-    def testSobolGpMean(self) -> None:
+    def test_SobolGpMean(self) -> None:
```
Balandat, line 112: nit...
```suggestion
    def test_SobolGPMean(self) -> None:
```

ax/utils/sensitivity/tests/test_sensitivity.py
```diff
@@ -256,7 +256,7 @@ def testSobolGpMean(self) -> None:
                         for j, col in enumerate(ind_dict[row]):
                             self.assertAlmostEqual(ind_dict[row][col], ind_tnsr[i, j])
 
-    def testSobolGpSampling(self) -> None:
+    def test_SobolGpSampling(self) -> None:
```
Balandat, line 259:
```suggestion
    def test_SobolGPSampling(self) -> None:
```

ax/utils/sensitivity/sobol_measures.py
```diff
@@ -423,8 +423,14 @@ def __init__(
...
             with torch.no_grad():
-                p = checked_cast(GPyTorchPosterior, self.model.posterior(x))
-            return link_function(p.mean, p.variance)
+                means, variances = [], []
+                # Since we're only looking at mean & variance, we can freely
+                # use mini-batches.
+                for x_split in x.split(split_size=32):
+                    p = checked_cast(GPyTorchPosterior, self.model.posterior(x_split))
+                    means.append(p.mean)
+                    variances.append(p.variance)
```
Balandat, line 432: Can we make this split size configurable as an input? 32 may be rather small in cases where memory isn't really an issue. Ideally we can even use some heuristic to determine this batching factor based on the type of model being used here (though that's a separate question beyond this PR).

Balandat (approved):

ax/utils/sensitivity/sobol_measures.py
saitcakmak: Yeah, this is definitely on the conservative side. Bernie's change was using a much larger value and still not using excess amounts of memory. I can make it configurable and bump up the default a bit. This is just an `n x d` tensor evaluated with `no_grad`, so the memory usage will generally be quite low.

PR #1886, opened 2023-10-02
Support empty input/outcome transform class lists

saitcakmak:
Summary:
Before this diff, calling `_make_botorch_input/outcome_transform` with empty list of classes would've raised an `IndexError`.

Also updates the backwards-compatibility decoders to convert `None` transform objects to `None` rather than an empty list of classes.

Reviewed By: susanxia1006

Differential Revision: D49833563


saitcakmak:
committed

PR #1920, opened 2023-10-18
Update use cases of deprecated fixed noise models

saitcakmak:
Summary:
Updates use cases of fixed noise models that were deprecated in  https://github.com/pytorch/botorch/pull/2052 & elsewhere.

The only remaining usage is `botorch_modular_registry`. I updated these to point to the parent models, so that they'll be less likely to cause backwards compatibility issues when we delete the models for good.

Differential Revision: D50431137


saitcakmak:
pr is disconnected from diff

PR #1974, opened 2023-11-10
Fix & temporarily disable raytune tutorial, print tutorial execution errors & memory usage

saitcakmak:
The tutorials script has been updated to mimic the setup in the BoTorch script to print out the errors encountered while running the tutorials.

The raytune tutorial mysteriously fails even after fixing the deprecation errors, so we'll temporarily remove it from the CI. The early stopping tutorial fails trial execution both when run locally and with this diff, so that's also disabled temporarily.

The tutorial changes address the deprecation warning / error:
```
DeprecationWarning: `tune.report` is deprecated.
Use `ray.train.report` instead -- see the example below:

from ray import tune     ->     from ray import train
tune.report(metric=1)    ->     train.report({'metric': 1})
```
Also reduces # of trials from 30 to 10.

Test Plan:

Ran `python scripts/make_tutorials.py -w $(pwd) -e -s` locally and checked the outputs (including with added errors in the tutorials).


scripts/make_tutorials.py
```diff
@@ -162,32 +166,46 @@ def gen_tutorials(
...
+                continue
             tutorial_path = Path(paths["tutorial_path"])
             print("Executing tutorial {}".format(tid))
-            start_time = time.time()
+            start_time = time.monotonic()
 
-            # try / catch failures for now
-            # will re-raise at the end
+            # Try / catch failures for now. We will re-raise at the end.
+            # TODO: [T163244135] Speed up tutorials and reduce timeout limits.
```
saitcakmak, line 179:
```suggestion
```

PR #1978, opened 2023-11-10
Make GenerationStrategy a GenerationStrategyInterface

danielcohenlive:
Summary: This is a noop.  These methods will be used in the follow up.

Differential Revision: D51045172


saitcakmak:
abandoned

PR #2000, opened 2023-11-17
Stop running cron_pinned on a schedule

saitcakmak:
This fails more often than it succeeds, creating a lot of noise. Its main use case is to test against newly pinned BoTorch just before a new release, which we can support using manual runs.


esantorella (approved):
Thanks!

PR #2005, opened 2023-11-20
Use default dtype in Experiment.clone_with

saitcakmak:
Summary: Replaces `Data.from_multiple_data` with `self.default_data_constructor.from_multiple_data`, which will use `MapData.from_multiple_data` where applicable.

Reviewed By: Balandat

Differential Revision: D51333709


saitcakmak:
committed

Issue #746, opened 2021-12-07
[Feature Request] Specifying fixed features to `AxClient.get_next_trial` (to conduct contextual BO using Service API)


MattiaVerticchio:
Hi everyone. First of all, thanks for working on Ax! 😁

I have a setting where I’d like to optimize a function based on a certain context. The context is, in this case, the day of the week.

If I understood correctly, the only way to have contextual BO in Ax is by using the Developer API; setting the context as input covariates, and then generating a `new_trial` with `fixed_features`:

```python
gpei = Models.BOTORCH(
    experiment=exp,
    data=exp.fetch_data()
)

generator_run = gpei.gen(n=1)

fixed_features = ObservationFeatures(parameters={"day_of_week": 0})

trial = exp.new_trial(
    generator_run=generator_run,
    fixed_features=fixed_features)
)
```

For this kind of task, the Developer API is more flexible but adds a lot of complexity. Is there any other way to do contextual BO, or pass `fixed_features` when using the simpler Service API?

Thanks!

lena-kashtelyan:
Hi @MattiaVerticchio, I think right now the Service API does not support this case, but it definitely could –– it seems reasonable for `get_next_trial` to be able to accept fixed features. I'll mark this as a feature request! 

In the meantime, you can hack around this like in this comment: https://github.com/facebook/Ax/issues/733#issue-1066610439, so basically generating the parameterizations outside of `AxClient` via `Models.BOTORCH` and then attaching them via `attach_trial`. Generally we really don't recommend this approach, but while the functionality you'd like isn't supported, you can make use of this perhaps?

lena-kashtelyan:
Putting this on wishlist for now, but we will likely support this soon!

sgbaird:
@MattiaVerticchio, do you mind including a reproducer for this example if you have one available?

Issue #566, opened 2021-04-23
Wishlist: Master Issue


lena-kashtelyan:
Feature requests marked as 'wishlist' will be gathered here going forward to improve discoverability of issues that report bugs or ask questions.

## Status: will likely be addressed in the long-term

- Hierarchical search spaces (https://github.com/facebook/Ax/issues/140)
- Optimization algorithm, suitable for combinatorial search spaces in Ax (https://github.com/facebook/Ax/issues/477)
- Include performance benchmarks in Ax documentation beyond the BoTorch benchmarks published in https://arxiv.org/abs/1910.06403 (https://github.com/facebook/Ax/issues/496)
- Exact equality parameter constraints (https://github.com/facebook/Ax/issues/510)

## Status: uncertain

- Nonlinear parameter constraints (https://github.com/facebook/Ax/issues/153)
- Adding "projects" to group Ax experiments (https://github.com/facebook/Ax/issues/189)
- Returning optimal ranges rather than single values for best parameters (https://github.com/facebook/Ax/issues/320)
- Ax on ARM / Raspberry pie (https://github.com/facebook/Ax/issues/412, blocked on official PyTorch support for ARM)
- ALEBO support for parameter constraints (https://github.com/facebook/Ax/issues/424)
- Selective noise inference for some observations in a field experiment (https://github.com/facebook/Ax/issues/491)
- Off-the-shelf models that do not use Gaussian noise (https://github.com/facebook/Ax/issues/491)

## Suggestions for setup changes:

- Migration of website to Docusaurus 2 (https://github.com/facebook/Ax/issues/466, status: blocked)

rpanai:
@lena-kashtelyan I already built a conda recipe see [PR](https://github.com/conda-forge/staged-recipes/pull/15371) whenever we fix the windows compatibility I'll let you know.

rpanai:
@lena-kashtelyan it is now available on conda-forge https://anaconda.org/conda-forge/ax-platform

lena-kashtelyan:
Wow, thank you so much, @rpanai!

CompRhys:
For "Exact equality parameter constraints" is the blocking factor an API design choice or a bandwidth choice? If it's bandwidth I am happy to take it on

CompRhys:
[...]

Issue #1951, opened 2023-11-02
[Feature Request] Support fixing parameter values in `get_next_trial()` in ServiceAPI 


Balandat:
Using `fixed_features` in [`Modelbridge.gen()`](https://github.com/facebook/Ax/blob/main/ax/modelbridge/base.py#L748) allows one to fix the values of certain parameters. This is useful in certain "contextual" optimization setups, see e.g. #1931 for a discussion. 

Currently, this is not supported in the Service API (i.e. in [`AxClient.get_next_trial()`](https://github.com/facebook/Ax/blob/main/ax/service/ax_client.py#L513)). If we want the Service API to be more powerful / the primary way of using Ax for most people, then we should consider adding support for this there.

sgbaird:
This would be a big benefit to me!
Related:
- https://github.com/facebook/Ax/issues/905
- https://github.com/facebook/Ax/issues/746
- https://github.com/facebook/Ax/issues/772
- https://github.com/facebook/Ax/issues/268

chanansh:
was this fixed? I cannot see it in the Service API.

Balandat:
cc @mpolson64

PR #2015, opened 2023-11-27
Support fixed features in Service API

AngelFP:
Addresses https://github.com/facebook/Ax/issues/746 (also in the wishlist https://github.com/facebook/Ax/issues/566).

As the title implies, this PR adds the possibility of specifying some `ObservationFeatures` as `fixed_features` in `AxClient.get_next_trial` and `AxClient.get_next_trials`. From my understanding, this is currently only possible with the developer API.

This is an option we would like to expose in [optimas](https://github.com/optimas-org/optimas), where we use the Service API for most of the BO generators.

### DIFF
--- a/ax/service/ax_client.py
+++ b/ax/service/ax_client.py
@@ -509,7 +509,10 @@ class AxClient(WithDBSettingsBase, BestPointMixin, InstantiationBase):
         wrap_error_message_in=CHOLESKY_ERROR_ANNOTATION,
     )
     def get_next_trial(
-        self, ttl_seconds: Optional[int] = None, force: bool = False
+        self,
+        ttl_seconds: Optional[int] = None,
+        force: bool = False,
+        fixed_features: Optional[ObservationFeatures] = None,
     ) -> Tuple[TParameterization, int]:
         """
         Generate trial with the next set of parameters to try in the iteration process.
@@ -522,6 +525,9 @@ class AxClient(WithDBSettingsBase, BestPointMixin, InstantiationBase):
                 failed properly.
             force: If set to True, this function will bypass the global stopping
                 strategy's decision and generate a new trial anyway.
+            fixed_features: An ObservationFeatures object containing any
+                features that should be fixed at specified values during
+                generation.
 
         Returns:
             Tuple of trial parameterization, trial index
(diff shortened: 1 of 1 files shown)


AngelFP:
Hi all,

has there been any update on this? We're currently having to use a [custom subclass](https://github.com/optimas-org/optimas/blob/706f80af8fecc71455d15531e7ebdc58afa78adb/optimas/generators/ax/service/custom_ax.py) of `AxClient` in order to have this feature, so it would be great if it could be merged :D (it would also address https://github.com/facebook/Ax/issues/1951).

Feel free to let me know if there's anything I should change in how I implemented the feature.

sgbaird:
Seconded on @AngelFP's comment! I've been itching to implement a fixed feature option within https://honegumi.readthedocs.io/ to better serve the researchers in my circle (coming back to this because I just wrapped up a long conversation about it with a colleague), but it's going to be much more straightforward to do so with this PR merged.

@Cesar-Cardoso and co., does the Facebook Internal - Linter check need to be fixed? Non- meta employees can't [view details on this](https://github.com/facebook/Ax/pull/2015/checks?check_run_id=19231251410).

ax/service/ax_client.py
```diff
@@ -1765,6 +1787,8 @@ def _gen_new_generator_run(self, n: int = 1) -> GeneratorRun:
                 parameters={}, trial_index=self._get_last_completed_trial_index()
             )
         )
+        if fixed_features:
+            fixed_feats.update_features(fixed_features)
```
sgbaird, line 1791: @Cesar-Cardoso This is really the only thing that's changing. The rest is plumbing to be able to pass `fixed_features` down to `_gen_new_generator_run(...)`.

Cesar-Cardoso:
Hi folks! Apologies for the slow response from my part. Support for fixed features is being added now in https://github.com/facebook/Ax/pull/2372. Only difference being the use of `FixedFeatures` instead of `ObservationFeatures`, as `ObservationFeatures` isn't meant to be used as input for the API. We also had some internal discussion about using `TParameterization` directly instead of `FixedFeatures`, but decided against it.

AngelFP:
Hi @Cesar-Cardoso , great to see this done! Thanks for looking into it.

sgbaird:
I'll try to put a MWE up soon, but a starting point might be https://github.com/facebook/Ax/blob/18d08424c4a11ca8ed9a50b760c87826e886be22/ax/service/tests/test_ax_client.py#L2838-L2863

PR #2123, opened 2024-01-08
Fix device issues with botorch_mes.

rentruewang:
Previously specifying any `torch_device` other than `"cpu"` would cause `RuntimeError` in torch. This is due to a device mismatch in MES's implementation.

### DIFF
--- a/ax/models/torch/botorch_mes.py
+++ b/ax/models/torch/botorch_mes.py
@@ -127,7 +127,7 @@ class MaxValueEntropySearch(BotorchModel):
         )
         bounds_ = bounds_.transpose(0, 1)
 
-        candidate_set = torch.rand(candidate_size, bounds_.size(1))
+        candidate_set = torch.rand(candidate_size, bounds_.size(1), device=self.device)
         candidate_set = bounds_[0] + (bounds_[1] - bounds_[0]) * candidate_set
 
         target_fidelities = {


danielcohenlive:
Hi @rentruewang, thank you so much for contributing to Ax! Could you provide a way to repro the problem you were having so I can verify this change?

rentruewang:
@danielcohenlive Thanks for the quick response!
The code is roughly as follows:

```
cli = AxClient(
    GenerationStrategy([
        Models.SOBOL(num_trials=5),
        Models.GPMES(num_trials=-1, model_kwargs: {'torch_device': 'cuda'})
    ])
)

# Create experiment here

for _ in range(10):
    # get next trial
    # evaluate
    # complete trail
```

This would emit a torch `RuntimeError` on the sixth step in the for loop.

Version is `0.3.6`

### My code changes:
Originally the `candidate_set`
https://github.com/facebook/Ax/blob/5230a53c798a8adf62147094d9014fe1725b4aaa/ax/models/torch/botorch_mes.py#L130
is constructed with `torch.rand`, which defaults to "cpu", causing problem here:
https://github.com/facebook/Ax/blob/5230a53c798a8adf62147094d9014fe1725b4aaa/ax/models/torch/botorch_mes.py#L131C32-L131C32
because `bounds_` is constructed here https://github.com/facebook/Ax/blob/5230a53c798a8adf62147094d9014fe1725b4aaa/ax/models/torch/botorch_mes.py#L125-L127 and has `device=self.device`

which is set by `torch_device` iirc.

danielcohenlive:
@rentruewang Great!  I've verified this problem and that your change fixes it.  I'm going to import it, it'll be reviewed and it will be in our next release.

rentruewang:
@danielcohenlive Great news! I'm wondering if there is any news to change the implementation of `GPMES` and others to use `BOTORCH_MODULAR`? My understanding is `GPMES` is simply `BOTORCH_MODULAR(model_kwargs={"botorch_acqf_class"=qMaxValueEntropy})` (and `BOTORCH_MODULAR` does work on GPU.

danielcohenlive:
cc @Balandat who might know the answer to the above ^^

saitcakmak (approved):
Thanks for the fix!

>  I'm wondering if there is any planned change to the implementation of GPMES and others to use BOTORCH_MODULAR? My understanding is GPMES is simply BOTORCH_MODULAR(model_kwargs={"botorch_acqf_class": qMaxValueEntropy}) (and BOTORCH_MODULAR does work on GPU).

I haven't tested MES in particular but `BOTORCH_MODULAR(model_kwargs={"botorch_acqf_class": qMaxValueEntropy})` should work as you mentioned. We migrated all BO models we use internally to `BOTORCH_MODULAR` setup and we plan to deprecate the legacy versions in H1. Replacing `GPMES` with the `BOTORCH_MODULAR` version would be a part of this.

PR #2126, opened 2024-01-09
Fix path to mnist_train_nas.py in early stopping tutorial

danielcohenlive:
image


saitcakmak (approved):
Thanks for the fix! Let's make sure https://github.com/facebook/Ax/actions/runs/7468012344/job/20322635991?pr=2126 succeeds before landing.

PR #2268, opened 2024-03-13
Fix flaky test for sensitivity analysis

bletham:
Summary:
This fixes a flaky test for sensitivity analysis. Currently the QMC
integration done for computing Sobol indices is seeded for reproducibility.
That done for computing derivative-based sensitivity measures is not. This diff
changes that so that the derivative-based measures will also have seeded QMC
with the same seed, to match Sobol indices.

A side effect of this is that it fixes the flaky test, since the results are no
longer random based on the QMC points.

Reviewed By: bernardbeckerman

Differential Revision: D54855136


saitcakmak:
merged

Issue #2221, opened 2024-02-23
get_countour_plot() not plotting all trials


davifebba:
Hello, I'm learning Ax and playing with the tutorial code of the Service API, but I noticed that render(ax_client.get_contour_plot()) does not display all trials over the response surface, but displays a total of (number of trials -1) samples. How to display the full set of samples on the response surface?

 For example, running an optimization campaign for 10 trials:

from ax.service.ax_client import AxClient, ObjectiveProperties
from ax.utils.measurement.synthetic_functions import hartmann6, Branin
from ax.utils.notebook.plotting import init_notebook_plotting, render
from ax.modelbridge.registry import Models
from ax.modelbridge.generation_strategy import GenerationStep, GenerationStrategy
from ax.models.torch.botorch_modular.surrogate import Surrogate
from botorch.acquisition.monte_carlo import (
    qNoisyExpectedImprovement,
)
from botorch.models.gp_regression import FixedNoiseGP

import numpy as np
init_notebook_plotting()

gs = GenerationStrategy(
    steps=[
        GenerationStep(  # Initialization step
            # Which model to use for this step
            model=Models.SOBOL,
            # How many generator runs (each of which is then made a trial) to produce with this step
            num_trials=5,
            # How many trials generated from this step must be `COMPLETED` # before the next one
            min_trials_observed=5,
        ),
        GenerationStep(  # BayesOpt step
            model=Models.BOTORCH_MODULAR,
            # No limit on how many generator runs will be produced
            num_trials=-1,
            model_kwargs={  # Kwargs to pass to `BoTorchModel.__init__`
                "surrogate": Surrogate(FixedNoiseGP),
                "botorch_acqf_class": qNoisyExpectedImprovement,
            },
        ),
    ]
)

ax_client = AxClient(generation_strategy = gs)

ax_client.create_experiment(
    name="hartmann_test_experiment",
    parameters=[
        {
            "name": "x1",
            "type": "range",
            "bounds": [0.0, 1.0],
            "value_type": "float",  # Optional, defaults to inference from type of "bounds".
            "log_scale": False,     # Optional, defaults to False.
        },
        {
            "name": "x2",
            "type": "range",
            "bounds": [0.0, 1.0],
        },
        {
            "name": "x3",
            "type": "range",
            "bounds": [0.0, 1.0],
        },
        {
            "name": "x4",
            "type": "range",
            "bounds": [0.0, 1.0],
        },
        {
            "name": "x5",
            "type": "range",
[...]

PR #2305, opened 2024-04-01
Fix get_countour_plot() not plotting all trials

ligerlac:
This fixes a bug where get_countour_plot() plots all but the last trial. The corresponding issue is here: https://github.com/facebook/Ax/issues/2221.

### DIFF
--- a/ax/service/ax_client.py
+++ b/ax/service/ax_client.py
@@ -1032,6 +1032,7 @@ class AxClient(WithDBSettingsBase, BestPointMixin, InstantiationBase):
                 )
                 return plot_contour(
                     model=not_none(self.generation_strategy.model),
+                    generator_runs_dict=self.experiment.trials,
                     param_x=param_x,
                     param_y=param_y,
                     metric_name=metric_name,


ligerlac:
AxClient.get_contour_plot() plots the training data of the model of the generation strategy as 'in-sample':
https://github.com/facebook/Ax/blob/29671645247d34004f7aef67d64870c9d13b0a59/ax/plot/helper.py#L189
However, this does not necessarily contain all trials of the experiment. For example, in the code snippet provided in the original issue (https://github.com/facebook/Ax/issues/2221#issue-2151849012), the model of the generation strategy is updated when ax_client.get_next_trial() is called. Therefore, the last trial will not be displayed in the contour plot. 
Passing all trials as generator_runs_dict like so https://github.com/facebook/Ax/pull/2305/commits/3d2eec27050ec84deb062280db1176e2f02a2da6# leads to duplicated data points except for the last trial. Alternatively, only trials that are not already present in the training data could be passed. Both solutions, however, result in a plot where the colored contours of the model do not take all displayed data points into account, which might be counter-intuitive.
Another possible solution would be to fit the model one more time before making the plot, ensuring the model is up-to-date with all completed trials. Something similar is done when retrieving the model's predictions here: https://github.com/facebook/Ax/blob/29671645247d34004f7aef67d64870c9d13b0a59/ax/service/ax_client.py#L1186-L1189
But that might be unexpected behavior.
Maybe @mgarrard can comment?

PR #2349, opened 2024-04-11
bump minimum python version to 3.10

sdaulton:
Bump minmum python version to 3.10 since scientific python is no longer supporting py3.9: https://scientific-python.org/specs/spec-0000/


saitcakmak (approved):
Thanks! Can you make Python 3.12 testing into a backlog task with the pyzmq as the blocker?

PR #2384, opened 2024-04-19
Make sure TestCase.setUp is executed in tests

saitcakmak:
Summary: The logger & warning filters in `TestCase.setUp` are only effective if it gets executed. Many test definitions were overwriting `setUp` without a `super().setUp()` call, ignoring these filters.

Reviewed By: esantorella

Differential Revision: D56375193


saitcakmak:
merged

PR #2396, opened 2024-04-25
Call object_from_json from decoder when issubclass(_class, SerializationMixin)

mpolson64:
Summary: As titled. This solves a relatively common problem (which may crop up again in some recently planning changes) in which the init args of some serializable Ax object is itself a serializable Ax object

Differential Revision: D56356035


saitcakmak:
merged

Issue #2452, opened 2024-05-13
MOO Tutorial potentially incorrect model


basarkutukcu:
In MOO Tutorial notebook (https://ax.dev/tutorials/multiobjective_optimization.html), "get_MOO_EHVI" is imported as below but never used.

```
# Factory methods for creating multi-objective optimization modesl.
from ax.modelbridge.factory import get_MOO_EHVI, get_MOO_PAREGO
```

Instead Models.BOTORCH_MODULAR is used. Is this a mistake or intentional?

```
  ehvi_model = Models.BOTORCH_MODULAR(
      experiment=ehvi_experiment,
      data=ehvi_data,
  )
```

bernardbeckerman:
Thanks for pointing this out! I'll put up a change to rectify this : )

basarkutukcu:
I actually commented on the PR but I think it's hard to see the comment there. I am copying the comment below. @bernardbeckerman What do you think?

Is this the correct fix for the problem? I was expecting a change to use `get_MOO_EHVI` instead of `Models.BOTORCH_MODULAR` as following.

```
      ehvi_model = get_MOO_EHVI(
          experiment=ehvi_experiment,
          data=ehvi_data,
      )
```

bernardbeckerman:
This new BOTORCH_MODULAR model type replaced the legacy get_MOO_EHVI in https://github.com/facebook/Ax/pull/1982 and should be used in its place going forward. Please let me know if you have further questions!

PR #2462, opened 2024-05-15
remove unused import from MOO tutorial

bernardbeckerman:
Summary: Reported here: https://github.com/facebook/Ax/issues/2452

Reviewed By: mgarrard

Differential Revision: D57359677


basarkutukcu:
Is this the correct fix for the problem? I was expecting a change to use `get_MOO_EHVI` instead of `Models.BOTORCH_MODULAR` as following.

```
      ehvi_model = get_MOO_EHVI(
          experiment=ehvi_experiment,
          data=ehvi_data,
      )
```

PR #2580, opened 2024-07-15
Minor updates to AxClient.fit_model

saitcakmak:
Summary:
I came across `AxClient.fit_model` while working on something else. It appeared to be a method that was split off from `get_model_predictions`, with leftover error messages that did not fit well with it as a standalone non-protected method.

This diff clears up the error messages and simplifies the definition of `AxClient.fit_model`.

Differential Revision: D59778614


ax/service/ax_client.py
```diff
@@ -1197,9 +1198,10 @@ def get_model_predictions(
                 "argument to `get_model_predictions`."
```
esantorella, line 1198: A few lines above, it says that `AxClient` only supports one-arm trials. Is that saying that `AxClient` doesn't support `BatchTrial`s? I didn't think that was the case. See e.g. here: https://github.com/facebook/Ax/blob/95881f4d0cdfc6dba525785d3e1ad7c3ecd8acd8/ax/service/ax_client.py#L1735

ax/service/ax_client.py
```diff
@@ -1197,9 +1198,10 @@ def get_model_predictions(
...
             )
 
-        # Fit model to ensure:
-        # - model is instantiated if needed
-        # - any new completed trials are fit to the model
+        # Model update is normally tied to the GenerationStrategy.gen() call,
+        # which is called from get_next_trial(). In order to ensure that predictions
+        # can be performed without the need to call get_next_trial(), we update the
+        # model with all attached data.
```
esantorella, line 1204: Thanks, this is clearer.

ax/service/ax_client.py
```diff
@@ -1255,29 +1257,11 @@ def fit_model(self) -> None:
         If no model is yet available a new one is instantiated. This
         may be the case when get_next_trial() has never been called.
```
esantorella, line 1258:
```suggestion

        If no model is yet available, a new one is instantiated. This
        may be the case when `self.get_next_trial()` has never been called.
```

ax/service/tests/test_ax_client.py
```diff
@@ -2103,25 +2103,15 @@ def test_unnamed_experiment_snapshot(self) -> None:
...
-    @patch(
-        "ax.modelbridge.base.observations_from_data",
-        autospec=True,
-        return_value=([get_observation1(first_metric_name="branin")]),
-    )
-    @patch(
-        "ax.modelbridge.random.RandomModelBridge.get_training_data",
-        autospec=True,
-        return_value=([get_observation1(first_metric_name="branin")]),
-    )
```
esantorella, line 2115: Nice, thanks!

ax/service/ax_client.py
```diff
@@ -1255,29 +1257,11 @@ def fit_model(self) -> None:
...
-                )
-            self.generation_strategy._fit_current_model(data=None)
-            logger.info("Successfully instantiated a model for the first time.")
-
-        # Model update is normally tied to the GenerationStrategy.gen() call,
-        # which is called from get_next_trial(). In order to ensure that predictions
-        # can be performed without the need to call get_next_trial(), we update the
-        # model with all attached data. Note that this method keeps track of previously
-        # seen trials and will update the model if there is newly attached data.
+        self.generation_strategy._maybe_transition_to_next_node()
```
esantorella, line 1264: Why might we need to transition to the next node?

esantorella (commented):
This may just reflect my ignorance of this area of the codebase, but could you clear up support for `BatchTrial`s and add a comment explaining the use of `ax_client.generation_strategy._maybe_transition_to_next_node()`?

ax/service/ax_client.py
saitcakmak: I don't know about AxClient more generally but this seems to be a limitation of the `get_model_predictions` method. I'll clarify this in the docstring.

ax/service/ax_client.py
saitcakmak: This could go either way. With this call, we fit the model that would be used to generate the next trial. Without it, we would just re-train (with updated data) the model that was used to fit the last trial. I'll add a note about this to the docstring.

esantorella (approved):
Thanks!

PR #2599, opened 2024-07-24
Consolidate `observe_noise_stds` and `observe_noise_sd`

esantorella:
Summary:
Context: `observe_noise_stds` was intended to track whether noise is observed on a metric-by-metric level for multi-objective problems. It should be a dict if noise levels are observed for some metrics and not for others. It can also be boolean and be used for single-objective problems. Confusingly, `observe_noise_sd` can be used for single-objective problems. In practice, these are always boolean and identical, because we have neither problems with partially observed noise nor methods for partially observed noise. It's also confusing to keep track of which is present on which problems and what types it might have.

In the spirit of YAGNI and to facilitate bringing single-objective and multi-objective problems closer together, this PR removes `observe_noise_stds` and replaces it with with `observe_noise_sd` where appropriate. It is alwayas boolean.

If we wind up needing this, we can put it back in, but after more refactoring, the right design might look different.

Note: Similar to D60194654, breaks backward compatibility, including for storage.

Differential Revision: D60202753

### DIFF
--- a/ax/benchmark/benchmark_problem.py
+++ b/ax/benchmark/benchmark_problem.py
@@ -11,7 +11,7 @@
 # in the UI.
 
 import abc
-from typing import Any, Dict, List, Optional, Protocol, runtime_checkable, Type, Union
+from typing import Any, Dict, List, Optional, Protocol, runtime_checkable, Type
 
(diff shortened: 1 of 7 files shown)


saitcakmak:
abandoned

PR #2607, opened 2024-07-26
Get rid of surrogate problem subclasses

esantorella:
Summary: These are no longer doing anything distinctive. The proper way to check if a problem is using a surrogate is to check the type of its runner.

Differential Revision: D60285442


saitcakmak:
abandoned

PR #2647, opened 2024-08-08
Create non_status_quo_arms prop on trials

danielcohenlive:
Summary:
As per title.

I considered adding this on base_trial but as of now don't see the use for it on non batch trials, so that can be done later if necessary.

Differential Revision: D60933893


saitcakmak:
abandoned

PR #2669, opened 2024-08-17
Split `BenchmarkRunner` into `BenchmarkWithGroundTruthRunner` and `BenchmarkWithoutGroundTruthRunner`

esantorella:
Summary:
Context: `BenchmarkRunner` has some methods that should be implemented if and only if there is a ground truth and some that are the opposite. It is hard to implement a new `BenchmarkRunner` subclass correctly because without abstract methods in the base class, there won't necessarily be an immediate error for subclasses that don't implement these methods.

This PR:
* Adds subclasses of `BenchmarkRunner`, `BenchmarkWithGroundTruthRunner` and `BenchmarkWithoutGroundTruthRunner`, each of which has only abstract methods that must be overridden and concrete methods that shouldn't need to be.
* Makes `BotorchTestProblemRunner` and `SurrogateRunner` subclasses of `BenchmarkWithGroundTruthRunner`
* Simplifies syntax for `outcome_names`

Differential Revision: D61431979

### DIFF
--- a/ax/benchmark/runners/base.py
+++ b/ax/benchmark/runners/base.py
@@ -21,46 +21,9 @@ from torch import Tensor
 
 
 class BenchmarkRunner(Runner, ABC):
+    outcome_names: list[str]
 
-    @property
     @abstractmethod
-    def outcome_names(self) -> list[str]:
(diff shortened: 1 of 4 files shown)


saitcakmak:
abandoned

PR #2684, opened 2024-08-21
Simplify ax/analysis/plotly directory structure

saitcakmak:
Summary:
Having separate sub-directories for each analysis makes the directory structure & import paths overly verbose. It also necessitates adding an `__init__.py` file for each sub-directory, which is easy to miss and can lead to `module not found` errors elsewhere (see D61607199).

This diff eliminates the separate sub-directory for the parallel coordinates analysis.

Differential Revision: D61611408


Balandat:
This caused a tutorial failure in botorch: https://github.com/pytorch/botorch/actions/runs/10495950816/job/29075527236

saitcakmak:
By fixing one issue, it unearthed a different issue :). Being fixed in https://github.com/facebook/Ax/pull/2692

PR #2692, opened 2024-08-22
Fix WithDBSettingsBase to no longer require SQLAlchemy

mpolson64:
Summary:
D61338432 accidentally imported from Ax's sql storage module at the top level instead of within the try/except block.

This will fix the Nightly job for both Ax and Botorch

Reviewed By: Cesar-Cardoso

Differential Revision: D61662444


saitcakmak:
merged

PR #2707, opened 2024-08-23
[Website] Use mathjax svg renderer

CristianLara:
We identified an issue where mathjax would fail to render certain math symbols. This happens on MacOS versions Ventura and later (13+) because of a change in the OS-supplied STIX fonts. We fix this here by changing the renderer from HTML+CSS (which tries to use local STIX fonts) to SVG.

This is a follow up to https://github.com/pytorch/botorch/pull/2481 which solved the same problem on the botorch website. See that PR for more details

| **Before**    | Screenshot 2024-08-23 at 4 40 19 PM |
| -------- | ------- |
| **After**  | Screenshot 2024-08-23 at 4 40 40 PM |

### DIFF
--- a/website/siteConfig.js
+++ b/website/siteConfig.js
@@ -68,7 +68,7 @@ const siteConfig = {
     `${baseUrl}js/plotUtils.js`,
     'https://buttons.github.io/buttons.js',
     `${baseUrl}js/mathjax.js`,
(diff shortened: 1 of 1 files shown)


Balandat (approved):
Thanks!

PR #2806, opened 2024-10-01
Make parameters ordered in SearchSpace

esantorella:
Summary:
This PR: Changes `SearchSpace.parameters` from a `dict` to an `OrderedDict`.

Context: In benchmarking, the order of parameters matters because parameters can be converted to a tensor. In the current Python version, this should work even without using `OrderedDict` because dicts are insertion-ordered, but this isn't an officially supported language feature or guaranteed to last.

One downside is that by annotating the parameters as an `OrderedDict`, we can't also annotate them as an immutable `Mapping`. For now, this is fine, because the parameters are mutated, so we can't annotate them as immutable anyway.

Differential Revision: D63641392

### DIFF
--- a/ax/core/search_space.py
+++ b/ax/core/search_space.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 
 import math
 import warnings
+from collections import OrderedDict
 from collections.abc import Hashable, Mapping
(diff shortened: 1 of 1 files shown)


saitcakmak:
abandoned

PR #2840, opened 2024-10-08
deepcopy model_gen_kwargs

sdaulton:
Summary: this ensures we don't do any in-place modification of model_gen_kwargs

Reviewed By: saitcakmak

Differential Revision: D64064492


danielcohenlive:
So I can push the change through

PR #2849, opened 2024-10-09
Fix is_optimal sometimes being used for y_metric_name

mpolson64:
Summary:
When generating the plot the y_metric name is inferred based on the other columns expected to be. When is_optimal was added this needed to be updated as well.

Now, instead of inferring we're just going to supply the y metric name directly -- this should be less error prone.

Differential Revision: D64123183


saitcakmak:
merged

Issue #2519, opened 2024-06-13
BUG: MBM sums up MOO outputs when given a single objective acquisition function


saitcakmak:
If the generation strategy uses MBM with a single objective acquisition function on an MOO problem, the outputs are simply summed together in the acquisition function using a `ScalarizedPosteriorTransform`.

Discovered while investigating https://github.com/facebook/Ax/issues/2514

## Repro:
Notebook for Meta employees: N5489742

### Setup the problem using AxClient
```python
import random

from ax.modelbridge.generation_strategy import GenerationStep, GenerationStrategy
from ax.modelbridge.registry import Models
from ax.service.ax_client import AxClient, ObjectiveProperties
from botorch.acquisition.monte_carlo import qNoisyExpectedImprovement


generation_strategy = GenerationStrategy(
    steps=[
        GenerationStep(
            model=Models.SOBOL,
            num_trials=2,
            min_trials_observed=1,
        ),
        GenerationStep(
            model=Models.BOTORCH_MODULAR,
            num_trials=-1,
            model_kwargs={
                "botorch_acqf_class": qNoisyExpectedImprovement,
            },
        ),
    ]
)

ax_client = AxClient(generation_strategy=generation_strategy)

ax_client.create_experiment(
    name="test_experiment",
    parameters=[
        {
            "name": "x1",
            "type": "range",
            "bounds": [0.0, 1.0],
        },
        {
            "name": "x2",
            "type": "range",
            "bounds": [0.0, 1.0],
        },
    ],
    objectives={
        "a": ObjectiveProperties(
            minimize=False,
        ),
        "b": ObjectiveProperties(
            minimize=False,
        ),
    },
)

def evaluate(parameters):
    return {"a": (random.random(), 0.0), "b": (random.random(), 0.0)}


for i in range(5):
    parameterization, trial_index = ax_client.get_next_trial()
    ax_client.complete_trial(
        trial_index=trial_index, raw_data=evaluate(parameterization)
    )
```
This runs fine and generates candidates.

### Investigate arguments to acquisition function
```python
from unittest import mock
with mock.patch.object(qNoisyExpectedImprovement, "__init__", side_effect=Exception) as mock_acqf:
    parameterization, trial_index = ax_client.get_next_trial()
```
This will raise an exception. Ignore it and check kwargs.
```
mock_acqf.call_args.kwargs["posterior_transform"]
```
[...]

PR #2913, opened 2024-10-17
Validate that an MO acqf is used for MOO in MBM/Acquistion

saitcakmak:
Summary:
This diff adds a validation that botorch_acqf_class is an MO acqf when `TorchOptConfig.is_moo is True`. This should eliminate bugs like https://github.com/facebook/Ax/issues/2519, which can happen since the downstream code will otherwise assume SOO.

Note that this only solves MBM side of the bug. Legacy code will still have the buggy behavior.

Differential Revision: D64563992


saitcakmak:
abandoned

Issue #2790, opened 2024-09-26
[GENERAL SUPPORT]: SEBO with parameter constraints


souravdey94:
### Question

I am trying the predict chemical reaction rates in different solvent combinations. I want to use SEBO because the parameter space can contain upto 30 solvents and in most cases the there are only 3 to 4 important solvents. Since, it is a composition problem, I need to use parameter constraints. But SEBO with parameter constraint is not implemented in Ax. Can you suggest me a work around? 

I have added a code snippet of the generation strategy and experiment section.

### Please provide any relevant code snippet if applicable.

```shell
length = len(solvent_names_minus1)
    print('length', length)

    torch.manual_seed(12345)  # To always get the same Sobol points
    tkwargs = {
    "dtype": torch.double,
    "device": torch.device("cuda" if torch.cuda.is_available() else "cpu"),
    }

    target_point = torch.tensor([0 for _ in range(length)], **tkwargs)
    print('target_point', target_point)

    SURROGATE_CLASS = SaasFullyBayesianSingleTaskGP


ax_client.create_experiment(
        name="solventproject",
 
        
        parameters=[
            {
                "name": solvent_names_minus1[i],
                "type": "range",
                "bounds": [float(range_min_minus1[i]), float(range_max_minus1[i])],
                "value_type": "float",  # Optional, defaults to inference from type of "bounds".
                "log_scale": False,  # Optional, defaults to False.
            }
            for i in range(len(solvent_names_minus1))
        ],
        objectives={"blend_score": ObjectiveProperties(minimize=False)},
        parameter_constraints=[sum_str],  # Optional.
        outcome_constraints=["lnorm <= 0.00"],  # Optional.
    )


gs = GenerationStrategy(
    name="SEBO_L0",
    steps=[
     
        GenerationStep(  # BayesOpt step
            model=Models.BOTORCH_MODULAR,
            # No limit on how many generator runs will be produced
            num_trials=-1,
            model_kwargs={  # Kwargs to pass to `BoTorchModel.__init__`
                "surrogate": Surrogate(botorch_model_class=SURROGATE_CLASS),
                "acquisition_class": SEBOAcquisition,
                "botorch_acqf_class": qNoisyExpectedHypervolumeImprovement,
                "acquisition_options": {
                    "penalty": "L0_norm", # it can be L0_norm or L1_norm.
                    "target_point": target_point, 
                    "sparsity_threshold": length,
                },
            },
        )
    ]
)
```

### Code of Conduct

[...]

PR #2938, opened 2024-10-22
Add Inequality Constraints to SEBO

CompRhys:
Following https://github.com/pytorch/botorch/pull/2588 being merged there is now no reason to stop people using inequality constraints with SEBO. This PR removes the ValueError check and adjusts the tests accordingly.

Connected to https://github.com/facebook/Ax/issues/2790


ax/models/torch/tests/test_sebo.py
```diff
@@ -317,19 +317,22 @@ def test_optimize_l0(self, mock_optimize_acqf_homotopy: Mock) -> None:
...
-            )
+        acquisition.optimize(
+            n=2,
+            search_space_digest=self.search_space_digest,
+            inequality_constraints=self.inequality_constraints,
+            fixed_features=self.fixed_features,
+            rounding_func=self.rounding_func,
+            optimizer_options=self.optimizer_options,
+        )
+        args, kwargs = mock_optimize_acqf_homotopy.call_args
```
Balandat, line 328: The unused assignment may upset the linter otherwise. 

```suggestion
        kwargs = mock_optimize_acqf_homotopy.call_args.kwargs
```

Balandat (approved):
Thanks - if you don't mind fixing that one small linter nit the rest lgtm!

Balandat (commented):
Actually there are a couple of other lint failures as well: https://github.com/facebook/Ax/actions/runs/11468956837/job/31951921724?pr=2938

CompRhys:
Merged the changes from https://github.com/facebook/Ax/pull/2935 and then fixed the flake8 complaints. I just went ahead and fixed everything flake8 complained about in the repo rather than just the files I touched. 

So used to relying on pre-commit that I always forget to lint manually when there's not a hook.

Balandat (changes requested):
Actually, you will want to make sure that you take into account the inequality constraints when in `get_batch_initial_conditions()` when you call this inside `_optimize_with_homotopy()`

CompRhys:
The naive solution would be to set it to `None` if inequality constraints are given which would then cause it to default to the behavior in botorch's `gen_batch_initial_conditions`. Would that be sufficient to be an acceptable solution here? Whether this is okay depends on how important the additional sampling around the pareto front is to performance (ping @dme65 as this was changed in the last few days).

Adding something more involved would require a bit more thought.

souravdey94:
Is SEBO available with parameter constraints in the latest version of Ax?

CompRhys:
No this was closed as I did a force-push and it broke the connection to this repo. I hope to get this sorted this week.

CompRhys:
failing tests here are due to the PR now depending on https://github.com/pytorch/botorch/pull/2636

CompRhys:
@saitcakmak @Balandat, I think this is now ready to be upstreamed now the related botorch PRs were merged. It should make no difference to existing users of L0 SEBO but there is a difference to anyone who might be using L1 SEBO as I moved the initial candidate selection based on the Pareto front to apply to both variants. I haven't benchmarked that change but it seemed to me that if was strange to not give already sparse solutions to the L1 optimizer if doing so was deemed necessary to getting good performance for the L0.

CompRhys:
@saitcakmak I would really appreciate it if this can be looked at/merged before the holidays, the request change is out of date now and can be cleared based on the work done in https://github.com/pytorch/botorch/pull/2636 such that we can make use of the hit-and-run polytope sampling so that the random samples satisfy the constraints

ax/models/torch/botorch_modular/sebo.py
```diff
@@ -233,16 +232,34 @@ def optimize(
...
+        # Prepare arguments for optimizer
+        optimizer_options_with_defaults = optimizer_argparse(
+            self.acqf,
+            optimizer_options=optimizer_options,
+            optimizer="optimize_acqf_homotopy",
+        )
+
+        _tensorize = partial(torch.tensor, dtype=self.dtype, device=self.device)
+        ssd = search_space_digest
+        bounds = _tensorize(ssd.bounds).t()
```
Balandat, line 244: This doesn't seem to add any value
```suggestion
        bounds = _tensorize(search_space_digest.bounds).t()
```

ax/models/torch/botorch_modular/sebo.py
```diff
@@ -233,16 +232,34 @@ def optimize(
...
+
+        optimizer_options["batch_initial_conditions"] = get_batch_initial_conditions(
```
Balandat, line 246: What if `batch_initial_conditions` is already contained in `optimizer_options`?

Balandat:
Sorry about the delay here @CompRhys, I'm at NeurIPS and so it's been a bit hectic. Sait can help get this over the finish line.

ax/models/torch/botorch_modular/sebo.py
```diff
@@ -233,16 +232,34 @@ def optimize(
...
+            acq_function=self.acqf,
+            raw_samples=optimizer_options_with_defaults["raw_samples"],
+            inequality_constraints=inequality_constraints,
+            fixed_features=fixed_features,
+            X_pareto=self.acqf.X_baseline,
+            target_point=self.target_point,
+            bounds=bounds,
+            num_restarts=optimizer_options_with_defaults["num_restarts"],
+        )
```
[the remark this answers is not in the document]
CompRhys: The motivation to include pareto samples for initializing the L0 optimization in addition to the random samples is the same for the L1 case. It creates an unfair comparison between the L1 and L0 methods to have different init strategies. Per https://github.com/facebook/Ax/pull/2938#issuecomment-2541761382 happy to revert this if would rather keep the change more atomic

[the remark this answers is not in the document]
saitcakmak: Ah, I see. The two methods use different optimizer. It makes sense to keep the two consistent.

ax/models/torch/botorch_modular/sebo.py
saitcakmak, line 246: Should this be set on `optimizer_options_with_defaults`? We now construct `optimizer_options_with_defaults` but don't pass it down to the optimizers.

I think this will also address the pyre failure:
fbcode/ax/models/torch/botorch_modular/sebo.py:246:8 Undefined attribute [16]: Optional type has no attribute `__setitem__`.

```suggestion
        optimizer_options_with_defaults["batch_initial_conditions"] = get_batch_initial_conditions(
```

ax/models/torch/botorch_modular/sebo.py
```diff
@@ -233,16 +232,34 @@ def optimize(
...
+        _tensorize = partial(torch.tensor, dtype=self.dtype, device=self.device)
+        ssd = search_space_digest
+        bounds = _tensorize(ssd.bounds).t()
+
+        optimizer_options["batch_initial_conditions"] = get_batch_initial_conditions(
+            acq_function=self.acqf,
+            raw_samples=optimizer_options_with_defaults["raw_samples"],
+            inequality_constraints=inequality_constraints,
+            fixed_features=fixed_features,
+            X_pareto=self.acqf.X_baseline,
```
saitcakmak, line 251: We have a pyre failure. The union types in PyTorch don't play well with type checkers :(
fbcode/ax/models/torch/botorch_modular/sebo.py:251:12 Incompatible parameter type [6]: In call `get_batch_initial_conditions`, for argument `X_pareto`, expected `Tensor` but got `Union[Tensor, Module]`.

```suggestion
            X_pareto=assert_is_instance(self.acqf.X_baseline, Tensor),
```

ax/models/torch/botorch_modular/sebo.py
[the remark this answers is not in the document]
CompRhys: will revert and then happy to fix/change in another PR

ax/models/torch/botorch_modular/sebo.py
```diff
@@ -300,26 +295,34 @@ def _optimize_with_homotopy(
...
-        )
+
+        if "batch_initial_conditions" not in optimizer_options:
+            optimizer_options["batch_initial_conditions"] = (
+                get_batch_initial_conditions(
+                    acq_function=self.acqf,
+                    raw_samples=optimizer_options_with_defaults["raw_samples"],
+                    inequality_constraints=inequality_constraints,
+                    fixed_features=fixed_features,
+                    X_pareto=self.acqf.X_baseline,
```
saitcakmak, line 306: We still need to update this
```suggestion
                    X_pareto=assert_is_instance(self.acqf.X_baseline, Tensor),
```

ax/models/torch/botorch_modular/sebo.py
```diff
@@ -300,26 +295,34 @@ def _optimize_with_homotopy(
...
-            # pyre-fixme[6]: For 3rd argument expected `Tensor` but got
-            #  `Union[Tensor, Module]`.
-            X_pareto=self.acqf.X_baseline,
-            target_point=self.target_point,
-            bounds=bounds,
-            num_restarts=optimizer_options_with_defaults["num_restarts"],
-        )
+
+        if "batch_initial_conditions" not in optimizer_options:
+            optimizer_options["batch_initial_conditions"] = (
```
saitcakmak, line 300:
```suggestion
            optimizer_options_with_defaults["batch_initial_conditions"] = (
```

ax/models/torch/botorch_modular/sebo.py
```diff
@@ -300,26 +295,34 @@ def _optimize_with_homotopy(
...
             homotopy=homotopy,
             num_restarts=optimizer_options_with_defaults["num_restarts"],
             raw_samples=optimizer_options_with_defaults["raw_samples"],
+            inequality_constraints=inequality_constraints,
             post_processing_func=rounding_func,
             fixed_features=fixed_features,
-            batch_initial_conditions=batch_initial_conditions,
+            batch_initial_conditions=optimizer_options_with_defaults[
+                "batch_initial_conditions"
+            ],
```
saitcakmak, line 325: Can you also update the tests to make sure this is passed down correctly?

saitcakmak (changes requested):

ax/models/torch/botorch_modular/sebo.py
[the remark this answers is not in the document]
CompRhys: > Ah, I see. The two methods use different optimizer. It makes sense to keep the two consistent.

sorry didn't see this reply before deciding it would be better to limit scope as otherwise more benchmarking might be needed. I have only tested the L0 implementation in my own work.

ax/models/torch/botorch_modular/sebo.py
CompRhys: For this is looks like there is a pyre pre-commit so will look at adding it to the pre-commit hooks https://github.com/facebook/pyre-check

PR #3131, opened 2024-11-29
Unpin SQLAlchemy dependency

Balandat:
As per #3008 this shouldn't be necessary anymore. Removing this pin as This is very restrictive and can cause problem for users' environments.


Balandat:
@paschai, @lena-kashtelyan not sure if this was the only thing holding back unpinning this but if it was we should make this change here.

PR #3153, opened 2024-12-06
Vizier sklearn transforms and PowerTransformY refactor

CompRhys:
This PR implements the Vizier Transforms as sklearn transformers and then creates a base class for applying such transformers. This refactor allows us to reuse most of the code in the PowerTransformY class and is hopefully fairly extensible. In future in may be possible to upstream the implementations of the vizier sklearn transformers somewhere else.

Because the power transformer already relied on sklearn this felt like a nice pattern. 

Still need to write tests for the Ax transformer classes but have written tests and double checked the behaviour of the transforms.


CompRhys:
closing as abandoned.

PR #3306, opened 2025-02-04
[GHA] Creating a new version only tracks relevant files to git

CristianLara:
Fixes an issue where unrelated files were getting added to the `docusaurus-versions` branch. In particular the changes from running tutorials (the tutorials themselves, the dbs) were getting persisted and causing problems for future runs of the tutorials.

### DIFF
--- a/.github/workflows/publish_website.yml
+++ b/.github/workflows/publish_website.yml
@@ -75,7 +75,7 @@ jobs:
         yarn
         yarn docusaurus docs:version ${{ inputs.new_version }}
 
-        git add --all
+        git add versioned_docs/ versioned_sidebars/ versions.json
         git commit -m "Create version ${{ inputs.new_version }} of site in Docusaurus"
         git push --force origin HEAD:docusaurus-versions
     - name: Build website


CristianLara:
Couple more things will need to happen to fully fix this:
1. Remove the accidentally committed files from `docusaurus-versions`, PR for this coming soon~
2. Propagate this change to `docusaurus-versions` branch by catching it up with main once this lands.

.github/workflows/publish_website.yml
```diff
@@ -75,7 +75,7 @@ jobs:
         yarn
         yarn docusaurus docs:version ${{ inputs.new_version }}
 
-        git add --all
+        git add versioned_docs/ versioned_sidebars/ versions.json
```
CristianLara, line 78: could have also just done `git add .` since these are the only modified files in the `website/` directory... but after the previous bug I'd rather be as specific as possible

saitcakmak: And thanks to the merge in l42, we should always be up to date with main for the rest of the repo, so we won't fall behind anywhere else. Versioned docs are only generated as part of this workflow, so we don't have to worry about merge conflicts either.

saitcakmak (approved):

PR #3329, opened 2025-02-07
[Docusaurus] Be more specific about archived versions

CristianLara:
The upgrade website has been released so we know exactly which versions are only available in the archive. Also `0.x.x` is not accurate anyways


saitcakmak (approved):
Shouldn't this be `<= 0.4.3`? Did those versions not publish?

CristianLara:
Thanks @saitcakmak you're correct, I was off by one. Fix for the same issue in botorch: https://github.com/pytorch/botorch/pull/2742

PR #3475, opened 2025-03-06
[Docusaurus] Hide recipes navbar item on unsupported doc versions

CristianLara:
Fix broken website build: https://github.com/facebook/Ax/actions/runs/13691586221/job/38285664204

Unfortunately the navbar items are not included in Docusaurus's versioning system. This is causing Docusaurus to attempt to show the newly added Recipes navbar item even when viewing the 0.5.0 version (which does not contain recipes) leading to a build failure.

Conditionally adding/removing navbar items is not natively supported by Docusaurus but they do provide a way for us to use our own custom navbar items. Our solution is to create a custom wrapper around the `DocSidebarNavbarItem` that only renders it if the current version is not in a provided list of versions.

Inspired by the official recommendation in this docusaurus github issue: https://github.com/facebook/docusaurus/issues/7227

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -96,10 +96,11 @@ module.exports={
           "position": "left"
         },
         {
-          "type": "docSidebar",
+          "type": "custom-docSidebar",
           "sidebarId": "recipes",
           "label": "Recipes",
-          "position": "left"
+          "position": "left",
+          "ignoreVersions": ["0.5.0"],
         },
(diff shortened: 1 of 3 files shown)


Balandat:
Thanks for fixing this. This seems like a pretty bad bug in Docusaurus if it breaks just b/c you add a new navbar item? Surprised this hasn't been fixed given that this has been known for 3 years now.

PR #3512, opened 2025-03-17
[Tutorials] Fix images for Docusaurus

CristianLara:
# Images by path

In the old tutorials we included images as base64 attachments but in the new ones we are specifying them via filepath. Relative filepaths in the notebook were breaking during the conversion process because the tutorial MDX file ends up in a different filesystem location while the image file was being left behind. We fix this by copying the image file to the tutorials docs location and updating the relative file path as appropriate.

A half-baked implementation of this was already present in the script but was being bypassed.

# Images as base64 attachments

The old tutorials were including images using base64 attachments which were not properly supported by the conversion script.

These image attachments are stored in the notebook cell's "attachments" field in base64 format with their associated mime_type and referenced in the markdown via attachment name.

The pattern we search for in the Markdown is `alt_text` with three groups:

- group 1 = alt_text (optional)
- group 2 = attachment_name
- group 3 = title (optional)

To represent this in MD we replace the attachment reference with the base64 encoded string as `{alt_text}`

This fix won't automatically propogate to the broken old tutorials, in a separate commit I'll fix them by updating the pre-built tutorial mdx stored in the `docusaurus-versions` branch


CristianLara:
PR to apply this fix to the versioned pre-built tutorials: https://github.com/facebook/Ax/pull/3513

Issue #3641, opened 2025-04-15
[Bug]: Pretty much the whole documentation is broken


NormanTUD:
### What happened?

Hi, 

I love using ax, but the documentation makes it REALLY hard. Almost all links are broken. Just one example:

https://ax.dev/docs/storage/

`If you add a custom Metric or Runner and want to ensure it is saved to JSON properly, create a RegistryBundle, which bundles together encoding and decoding logic for use in the various save/load functions as follows:`

Clicking on all these 3 links gives me links like https://ax.readthedocs.io/en/latest/core.html#module-ax.core.metric , but none of them work.

Its basically the same for google. I know that indexing a new site takes a long time, but it became pretty much impossible to find anything in the documentation since the changes for the documentation system.

Please run something like linkchecker on it, and try to fix the broken links. It would be much appreciated.

### Please provide a minimal, reproducible example of the unexpected behavior.

Clicking through the documentation, trying to click some links, like half of them fail.

### Please paste any relevant traceback/logs produced by the example provided.

```shell

```

### Ax Version

0.5.0

### Python Version

Doesn't matter here

### Operating System

Linux

### Code of Conduct

- [x] I agree to follow Ax's Code of Conduct

Balandat:
Seems like most of this is just not there anymore? #3552  

@mpolson64, @CristianLara

Cesar-Cardoso:
Thanks for reporting this! A few things going on that lead to this:
1. In https://github.com/facebook/Ax/pull/3552 we removed the Sphinx documentation for these classes. You can read more about it in the pull request.
2. We haven't yet published a new version of the Ax website without references to these classes' deleted documentation.
3. The Ax website links are pointing to the `latest` version of our documentation rather than `stable`. This lead to all the broken links you see. We're fixing that in https://github.com/facebook/Ax/pull/3643.

TL;DR: we're landing the fix these broken links right now, but we're removing them from the website in the near future.

Balandat:
@NormanTUD I'd be curious to hear from you about the value of having a full API reference in readthedocs vs. only having that for the "user facing" API and directing more advanced users (such as you) to the docstrings in the code.

NormanTUD:
Ah I see. But still, it would be greatly appreciated by many users (including me) to have a full reference online that is easily googleable.

Thanks for calling me an advanced user, even though I don't really consider myself that 😅

[...]

PR #3643, opened 2025-04-15
Docusaurus: stable docusaurus links to stable ReadTheDocs

CristianLara:
These were erroneously pointing to the latest version in ReadTheDocs. We changed the API docs in RTD (reflected on version "latest") which broke these links from the stable docs in Docusaurus.

When making retroactive changes to the branch `docusaurus-versions` we must ensure that it won't create merge conflicts with the `main` branch. In this case that means only updating the links in `website/versioned_docs`. This branch still has erroneous links in the top level `docs/` directory but those get overwritten by the latest changes in `main` when deploying the website.

Fixes issue reported in https://github.com/facebook/Ax/issues/3641


mpolson64 (approved):
lgtm

Issue #3659, opened 2025-04-17
[Bug]: Typo/inconsistency between data_loader_config and error messages mentioning data_loader_options.


CompRhys:
### What happened?

```
[WARNING 04-17 15:05:27] ax.modelbridge.base: `fit_out_of_design` is deprecated. Please pass as `data_loader_options.fit_out_of_design` instead.
```

In the code there are no references to `data_loader_options` outside of docstrings/warnings/function-names. Should these not be updated to `data_loader_config`? It requires some reading of the code to understand how to change over to the newer inputs.

It would also be nice to be able to give the config as a dict `data_loader_config": {"fit_out_of_design": False}` when instantiating a `GenerationStrategy` rather than needing to define the object like `data_loader_config": DataLoaderConfig(**{"fit_out_of_design": False})`

cc @SebastianAment

### Please provide a minimal, reproducible example of the unexpected behavior.

-

### Please paste any relevant traceback/logs produced by the example provided.

```shell

```

### Ax Version

github main

### Python Version

3.12

### Operating System

MacOS

### Pull Request

None

### Code of Conduct

- [x] I agree to follow Ax's Code of Conduct

PR #3663, opened 2025-04-18
Update modelbridge configs from `data_loader_options` to `data_loader_config`

Cesar-Cardoso:
Summary: See https://github.com/facebook/Ax/issues/3659

Differential Revision: D73263081

### DIFF
--- a/ax/modelbridge/torch.py
+++ b/ax/modelbridge/torch.py
@@ -120,7 +120,8 @@ class TorchAdapter(Adapter):
                 `model_gen_options` passed to the `Adapter.gen` method.
             torch_device: The device to use for any torch tensors and operations
                 on these tensors.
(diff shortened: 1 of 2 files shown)


ax/modelbridge/torch.py
```diff
@@ -120,7 +120,7 @@ def __init__(
                 `model_gen_options` passed to the `Adapter.gen` method.
             torch_device: The device to use for any torch tensors and operations
                 on these tensors.
-            data_loader_options: A dictionary of options for loading data.
+            data_loader_config: A dictionary of options for loading data.
```
CompRhys, line 123: This also isn't a dictionary, it has to be an object which is what I was interested in per second comment of https://github.com/facebook/Ax/issues/3659

PR #3768, opened 2025-05-09
Rename unified.py to arm_effects.py

mgarrard:
Summary: This is the last diff in the stack that deletes old observed/predicted effects analyses, moves the utils to the shared utils file, and we want to rename unified to arm_effects for clarity

Differential Revision: D74506602


fivemtebex:
Very clean code with good test coverage. This is a great example of best practices.

PR #3895, opened 2025-06-07
Make model-registry-base easier to make child classes

CompRhys:
If I would like to reuse the machinery of `ModelRegistryBase` I need to manipulate the Ax Global `MODEL_KEY_TO_MODEL_SETUP`. This doesn't feel nice to me and so I propose to make a property in `ModelRegistryBase` to return `MODEL_KEY_TO_MODEL_SETUP` and then in any `MyGenerators` child Enum I can just make my own `MY_MODEL_KEY_TO_MY_MODEL_SETUP` and overwrite that property and in principle everything works and Ax is easier to extend.


saitcakmak (approved):
This looks reasonable to me. 

For added content: We're not really happy with the registry setup as is, and we may get rid of it completely or update it to return `GeneratorSpec` objects in the future. The overall feeling is that there are currently too many layers of objects involved in specifying a `GenerationStrategy`, and registry in particular seems completely redundant to `GeneratorSpec`.

CompRhys:
The main API thing that causes me to get a bit lost is I am not sure why both nodes and steps are different things in generation strategies. I think these should be merged but I also to use the `ExternalGenerationNode` a fair amount for example I have a TurBO-type external generation node that I am not sure would be easy to get to live inside a `Generator`/model

saitcakmak:
> why both nodes and steps are different things in generation strategies. I think these should be merged 

I totally agree with this. Cleaning up the distinction between the two is one of the backlog tasks we have.

> I also to use the ExternalGenerationNode a fair amount for example I have a TurBO-type external generation node that I am not sure would be easy to get to live inside a Generator/model

We have an internal implementation of TRBO / MORBO as a subclass of `TorchGenerator`, but it is too messy to share externally. In that case, the `TorchAdapter` just handles the data processing and the custom `Generator` class operates TRBO with the tensor inputs. 

I generally think `ExternalGenerationNode` has its place. We have an upcoming Ax paper in the AutoML conference. We've used it to run all the external baseline methods for that paper. For methods that don't fit into the typical `Adapter/Generator` setup, EGN is a nice alternative that offers a lot of flexibility, and you can combine it with other nodes in `GenerationStrategy` as needed as well.

saitcakmak:
Hi @CompRhys. Looks like there is a merge conflict with the main branch and I don't have write permissions. Can you rebase this onto main?

CompRhys:
merged in upstream main but didn't need to merge anything so not sure if that will fix for your side.

saitcakmak:
Thanks! Our internal checks for merge conflicts can sometimes be a bit strict and flag things that can be handled automatically on rebase. I'll re-import and see if anything complains.

PR #3898, opened 2025-06-09
[GHA] Always create new docusaurus version, only keep latest patch fo…

CristianLara:
…r given major.minor version

Before this change we would only create new versions in docusaurus if the major or minor version changed, but this prevents the website from showing documentation changes bundled with patch releases.

The change here is to always create a new version in docusaurus, but ensure that for a given X.Y major.minor version we only track the latest patch. We accomplish this by deleting versions we previously created in docusaurus if they only differ by the patch version (i.e. `Z` in `X.Y.Z`)

While Docusaurus provides an npm command to easily create new versions, deleting them has to be done manually using the instructions provided here (which I've codified as a step in the workflow): https://docusaurus.io/docs/versioning#deleting-an-existing-version

Test Plan:

- Applied these same changes to my fork in https://github.com/CristianLara/Ax/pull/25
- Then created a new release on my fork to test the workflow
    - The final step of the website build failed due to unrelated broken links (fork is quite old): https://github.com/CristianLara/Ax/actions/runs/15542889463/job/43757621741
    - But I confirmed that the commit it created properly deleted the related files: https://github.com/CristianLara/Ax/commit/cd54b42773c92330c908b90a814e50077a233f59

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -22,39 +22,8 @@ jobs:
       pinned_botorch: true
     secrets: inherit
 
-  check-versions:
-    needs: tests-and-coverage-pinned
-    name: Check if major or minor version changed
-    runs-on: ubuntu-latest
-    outputs:
-      major_minor_changed: ${{ steps.compare.outputs.major_minor_changed }}
-    steps:
-    - uses: actions/checkout@v4
-      with:
-        fetch-depth: 0
-        fetch-tags: true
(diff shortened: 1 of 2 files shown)


CristianLara:
Will open a PR with the same change in botorch shortly

CristianLara:
> Will open a PR with the same change in botorch shortly

https://github.com/pytorch/botorch/pull/2874

PR #4134, opened 2025-08-12
clone_reset support for DeterministicGenerationNode

hvarfner:
Summary:
clone_reset currently does not work for DeterministicGenerationNode, as the iterator is not reset when the method is called on the generation strategy.

This causes issues unittesting benchmarking functionality, such as benchmark_replication.

Rollback Plan:

Differential Revision: D80093006


saitcakmak:
abandoned

PR #4198, opened 2025-08-25
[Website][Tutorials] Handle multiple image paths in same md doc

CristianLara:
The ipynb to mdx conversion script has a step to copy images to the mdx destination and update the relative image paths. The issue was that modifying the md buffer causes the static regex match indices to become out of date.

The fix here here is to process the matches in reverse order, that way modifications at the end of the buffer don't affect the indices for the next replacements (which are earlier in the buffer). This is the same solution used in the `handle_image_attachments()` function which experienced a similar problem

@eonofrey found this issue while adding images to a tutorial in f2186fd9fbbfc403c443e80d7cbc5011f9bb376b

| Before    | After |
| -------- | ------- |
|        image         |     image       |

### DIFF
--- a/scripts/convert_ipynb_to_mdx.py
+++ b/scripts/convert_ipynb_to_mdx.py
@@ -282,22 +282,20 @@ def handle_image_paths_found_in_markdown(
     if not searches:
         return markdown
 
-    # Convert the given Markdown to a list so we can delete the old path with the new
-    # standard path.
-    markdown_list = list(markdown)
-    for search in searches:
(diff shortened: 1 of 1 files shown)


CristianLara:
will open same PR for botorch

edit: https://github.com/pytorch/botorch/pull/2992

CristianLara:
I should have caught this when making the same fix to `handle_image_attachments()` in f3119ad4cdf7c7d6b22f8fb5db7e76a20b378793

PR #4217, opened 2025-08-27
Expand model space for ChoiceParameters

sunnyshen321:
Summary: Update `set_model_space` to be able to expand ChoiceParameter as needed to cover the values found in arm_data

Differential Revision: D81179672


saitcakmak:
Replaced with https://github.com/facebook/Ax/pull/5210

PR #4246, opened 2025-09-04
Make model based Derelativize.transform_optimization_config no-op if a fitted model is not yet available

saitcakmak:
Summary:
In order to use model predicted status quo values, we first need to have a fitted model. This prevents transforming optimization config in `Adapter._transform_data` and using it when instantiating the subsequent transforms, which we want to do to avoid the need to separately derelativize the opt config in transforms like `Winsorize` & `BilogY` (WIP D81597506).

This diff changes the behavior of `Derelativize.transform_optimization_config` to be no-op rather than error out when a fitted model is not yet available. This will leave the `optimization_config` unchanged in `Adapter._transform_data` and untransform it using the model predictions in `Adapter.gen`.

Reviewed By: dme65

Differential Revision: D81624306

### DIFF
--- a/ax/adapter/transforms/derelativize.py
+++ b/ax/adapter/transforms/derelativize.py
@@ -34,13 +34,15 @@ class Derelativize(Transform):
     """Changes relative constraints to not-relative constraints using a plug-in
     estimate of the status quo value.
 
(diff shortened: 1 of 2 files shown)


saitcakmak:
abandoned

PR #4550, opened 2025-11-18
Fix typo in automating.ipynb documentation

delfoxav:
double "reading" removed


paschai:
Thanks for putting up this correction - we have merged this internally and you will see these changes reflected in the next version!

PR #4551, opened 2025-11-18
Fix step numbering in automating.ipynb

delfoxav:
- corrected heading level of step 2 to match the level of others steps.
- moved step 5 -> 4, 6 -> 5 as there were no step 5


paschai:
Thanks for putting up this correction - we have merged this internally and you will see these changes reflected in the next version!

Issue #4599, opened 2025-11-27
[Bug]: from_config.py Floating point modulo bug in RangeParameter step_size validation


wizardlyrocket8:
### What happened?

I am sorry this is my first time writing a bug report (At least I think it is a bug).
But I was trying to make a sum constraint from a list of RangeParameterConfig, so because the SumConstraint only takes RangeParameters, I decided to convert RangeParameterConfigs to RangeParameters.

            def get_client_constraints(parameters):
                param_range_list = [parameter_from_config(parameter) for parameter in parameters if parameter.name in ['[Fe3+] (g/L)','[Co2+] (g/L)', '[Cu2+] (g/L)','[K+] (g/L)']]
                sum_constraint = SumConstraint(param_range_list, is_upper_bound= True, bound = 100.0)
                return sum_constraint

This is where the problem occurred: "ax.exceptions.core.UserInputError: The range of the parameter must be evenly divisible by the step size."
I went to explore and found that because the bounds for parameters are [0.0, 100.0] and the step size is 0.1, I get the error from this if statement: 

            if (upper - lower) % step_size != 0:
                raise UserInputError(
                    "The range of the parameter must be evenly divisible by the "
                    "step size."
                )

### Please provide a minimal, reproducible example of the unexpected behavior.

print((100.0 - 0.0)%0.1)
0.09999999999999445

### Please paste any relevant traceback/logs produced by the example provided.

```shell

```

### Ax Version

1.2.1

### Python Version

3.13.9

### Operating System

MAC

### Pull Request

None

### Code of Conduct

- [x] I agree to follow Ax's Code of Conduct

Cesar-Cardoso:
Hello there! Thanks for reporting this issue. Indeed it was a bug! It should be fixed after https://github.com/facebook/Ax/pull/4604

PR #4910, opened 2026-02-16
Fix RangeParameterConfig in from_config.py returns ChoiceParameter instead of RangeParameter #4599

wizardlyrocket8:
The original logic would return a ChoiceParameter after checking all the requirement for the RangeParameter. Now it checks if the instance is RangeParameterConfig and all of the conditions are met, the function return a RangeParameter.
    
if isinstance(config, RangeParameterConfig):
        lower, upper = config.bounds

        # TODO[mpolson64] Add support for RangeParameterConfig.step_size native to
        # RangeParameter instead of converting to ChoiceParameter
        if (step_size := config.step_size) is not None:
            if not (config.scaling == "linear" or config.scaling is None):
                raise UserInputError(
                    "Non-linear parameter scaling is not supported when using "
                    "step_size."
                )

            remainder = (upper - lower) % step_size
            # Use tolerance-based comparison to handle floating point precision issues
            if not np.isclose(remainder, 0) and not np.isclose(remainder, step_size):
                raise UserInputError(
                    "The range of the parameter must be evenly divisible by the "
                    "step size."
                )
          return ChoiceParameter(
              name=config.name,
              parameter_type=_parameter_type_converter(config.parameter_type),
              values=[*np.arange(lower, upper + step_size, step_size)],
              is_ordered=True,
              sort_values=False,  # already sorted by np.arange.
          )
        return RangeParameter(
            name=config.name,
            parameter_type=_parameter_type_converter(config.parameter_type),
            lower=lower,
            upper=upper,
            log_scale=config.scaling == "log",
        )

### DIFF
--- a/ax/api/utils/instantiation/from_config.py
+++ b/ax/api/utils/instantiation/from_config.py
@@ -50,21 +50,21 @@ def parameter_from_config(
                     "The range of the parameter must be evenly divisible by the "
                     "step size."
                 )
-
-            return ChoiceParameter(
+            return RangeParameter(
                 name=config.name,
                 parameter_type=_parameter_type_converter(config.parameter_type),
-                values=[*np.arange(lower, upper + step_size, step_size)],
-                is_ordered=True,
+                lower=lower,
+                upper=upper,
+                log_scale=config.scaling == "log",
             )
-
-        return RangeParameter(
+        return ChoiceParameter(
             name=config.name,
             parameter_type=_parameter_type_converter(config.parameter_type),
-            lower=lower,
-            upper=upper,
-            log_scale=config.scaling == "log",
+            values=[*np.arange(lower, upper + step_size, step_size)],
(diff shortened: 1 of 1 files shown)


ax/api/utils/instantiation/from_config.py
```diff
@@ -50,21 +50,21 @@ def parameter_from_config(
...
-        return RangeParameter(
+        return ChoiceParameter(
             name=config.name,
             parameter_type=_parameter_type_converter(config.parameter_type),
-            lower=lower,
-            upper=upper,
-            log_scale=config.scaling == "log",
+            values=[*np.arange(lower, upper + step_size, step_size)],
+            is_ordered=True,
         )
```
saitcakmak, line 65: Looks like the return values in & out of the if statement changed places here. This doesn't seem like the correct behavior here.

PR #4985, opened 2026-03-06
Enable autoescaping in report templates to prevent HTML injection

0xmrma:
## Summary
This PR enables Jinja2 autoescaping for Ax HTML report rendering and adjusts templates to preserve intended safe HTML insertion.

## Security impact
Without autoescape, user-controlled strings (e.g. experiment_name) can be rendered into HTML without escaping, enabling HTML injection (XSS) when exported HTML reports are opened/viewed.

## Changes
- Enable Jinja2 autoescape for HTML/XML templates in `ax/utils/report/render.py` using `select_autoescape(["html","xml"])`.
- Keep `html_elements` rendering behavior by explicitly marking elements as safe in `simple_template.html` (`{{ element | safe }}`).
- Mark `cell.html` as safe in `sufficient_statistic.html` to preserve existing behavior where `cell.html` is expected to contain trusted HTML.

## Test / verification
Added a lightweight smoke test (`scripts/test_report_autoescape_smoke.py`) that:
- loads `ax/utils/report/render.py` without importing the full Ax stack,
- renders a report with an HTML payload as experiment_name,
- asserts the payload is escaped.

Run:
```bash
python scripts/test_report_autoescape_smoke.py
```

## Notes
This is a defense-in-depth hardening. If `cell.html` can be influenced by untrusted input, it should be sanitized upstream before being passed into the report templates.

### DIFF
--- a/ax/utils/report/render.py
+++ b/ax/utils/report/render.py
@@ -10,7 +10,7 @@ import os
 import pkgutil
 
 from ax.plot.render import _js_requires, _load_css_resource as _load_plot_css_resource
-from jinja2 import Environment, FunctionLoader
+from jinja2 import Environment, FunctionLoader, select_autoescape
 
 REPORT_MODULE_NAME = "ax.utils.report"
 
@@ -142,4 +142,7 @@ def _load_html_template(name: str) -> str:
 
 
 def _get_jinja_environment() -> Environment:
-    return Environment(loader=FunctionLoader(_load_html_template))
+    return Environment(
+        loader=FunctionLoader(_load_html_template),
+        autoescape=select_autoescape(["html", "xml"]),
+    )
--- a/ax/utils/report/resources/simple_template.html
+++ b/ax/utils/report/resources/simple_template.html
@@ -4,6 +4,6 @@
 {% extends "base_template.html" %}
 {% block content %}
 {% for element in html_elements %}
-{{element}}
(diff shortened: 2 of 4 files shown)


0xmrma:
Hi maintainers — I signed the Meta CLA (please let me know if it hasn’t propagated yet).
This PR enables Jinja2 autoescaping for Ax HTML report templates and adds a small smoke test reproducing the original HTML injection (experiment_name) and verifying it’s now escaped.
I kept html_elements rendering as |safe (intended behavior: elements are pre-built HTML strings). I also kept cell.html as |safe to preserve existing behavior in sufficient_statistic.html, while relying on autoescape to protect other fields like titles/captions.
Happy to adjust if you prefer cell.html to be escaped by default and only marked safe by trusted callers.

0xmrma:
Hi maintainers — I signed the Meta CLA (confirmation received).
The “Meta CLA Check” is still showing “Missing CLA”. Could you please re-run/refresh that check on this PR when you have a moment? Thanks!

0xmrma:
Hi maintainers
Friendly ping on #4985 when you have time.
CLA is now marked valid and Meta Import Checks passed; only Import Status (Queued/not imported) + workflow approval remain.
Could you please approve the pending workflows / trigger the import so the PR can be reviewed?
Thank you!

PR #5211, opened 2026-05-28
Expose `digits` on `RangeParameterConfig`

### LINKED ISSUE #5108: [FEATURE REQUEST]: Allow ChoiceParameter with 1001 values
### Motivation

Hello Ax team.
My request is very simple: Please allow `ChoiceParameter` to accept 1001 values instead of 1000.
Why:
- An even range from 0 to 1 in 0.001 steps yields an array of 1001 elements.
- Or an even range of integers from 0 to 1000 has 1001 elements. (Just to make my point. I know that here one would use `RangeParameter(parameter_type="int")`)
This applies in general if one wants to evenly split an interval by 1/1000th of its range, which is very common in real-world experimenting.

Currently, Ax raises the error: "UserInputError: `ChoiceParameter` with more than 1000 values is not supported! Use a `RangeParameter` instead."

Thanks very much in advance.

### Describe the solution you'd like to see implemented in Ax.

Change `MAX_VALUES_CHOICE_PARAM = 1000` to `MAX_VALUES_CHOICE_PARAM = 1001` in `ax/core/parameter.py`.

### Describe any alternatives you've considered to the above solution.

_No response_

### Is this related to an existing issue in Ax or another repository? If so please include links to those Issues here.

_No response_

### Code of Conduct

saitcakmak:
## Summary
- Adds `digits: int | None = None` to `RangeParameterConfig`, plumbed through `parameter_from_config` to `RangeParameter`. Gives client-API users a first-class way to limit the resolution of a float range parameter without the `client._experiment.search_space.parameters[name].digits = N` workaround discussed in #5108.
- `digits` and `step_size` are mutually exclusive and serve different intents:
  - `step_size` materializes an ordered `ChoiceParameter` over an explicit grid; optimizer treats the parameter as discrete. Capped at 1000 grid points.
  - `digits` keeps the parameter as a continuous `RangeParameter` and rounds suggested values to N decimals at output time; optimizer treats it as continuous. Float-only, works with log/logit scaling, no cardinality cap.
- Both restrictions (`step_size` ∧ `digits` set; `digits` on non-float) raise `UserInputError` in `parameter_from_config`. Docstring on `RangeParameterConfig` documents the distinction.

Closes #5108.

## Test plan
- [x] New unit tests in `test_from_config.py` cover: float happy-path, log-scale happy-path, mutual-exclusion error, int+digits error.
- [x] `python -m pytest ax/api/utils/instantiation/tests/test_from_config.py -xvs` → 4 passed.
- [x] `ufmt` and `python -m flake8 -j 1` clean on all 3 changed files.

### DIFF
--- a/ax/api/configs.py
+++ b/ax/api/configs.py
@@ -18,6 +18,23 @@ class RangeParameterConfig:
     """
     Allows specifying a continuous dimension of an experiment's search space
     and internally validates the inputs.
+
+    ``step_size`` and ``digits`` are two distinct ways to express limited
+    resolution and are mutually exclusive:
+
+    - ``step_size`` materializes an ordered ``ChoiceParameter`` over the
+      explicit grid ``[lower, lower + step, ..., upper]``. The optimizer
+      treats the parameter as discrete (with nearest-neighbor acquisition
+      optimization for low-cardinality grids). Use this when the domain
+      *is* the grid -- e.g. a small set of hardware settings. Requires
+      linear scaling and ``(upper - lower) % step_size == 0``.
+    - ``digits`` keeps the parameter as a continuous ``RangeParameter``
+      and rounds suggested values to ``digits`` decimal places at output
+      time. The optimizer treats the parameter as continuous. Use this
+      when the underlying quantity is continuous but the
+      measurement/control resolution is limited -- e.g. a laser intensity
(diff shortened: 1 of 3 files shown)


saitcakmak:
Working on expanding core support for step_size parameter instead.

PR #5236, opened 2026-06-18
Enable Pyrefly for fbcode/ax (#5236)

maggiemoss:
Summary:

Migrates `fbcode/ax` from Pyre to Pyrefly by removing the root `python.set_pyrefly(False)` opt-out so the tree inherits `set_pyrefly(True)` from the fbcode root default. Pre-existing type errors are suppressed via annotation-only `# pyrefly: ignore` comments; no runtime changes.

___

Reviewed By: connernilsen, saitcakmak

Differential Revision: D109082535


connernilsen (approved):
Review automatically exported from Phabricator review in Meta.

PR #5249, opened 2026-07-02
Fix silent NaN results and KeyError in test of no effect

### LINKED ISSUE #5248: [Bug]: TestOfNoEffectAnalysis crashes with KeyError 'n' on standard Client data, and silently reports "no effect" for sem=0 or single-arm trials
### What happened?

`TestOfNoEffectAnalysis` (and the underlying functions in `ax/utils/stats/no_effects.py`) mishandles three inputs that are easy to produce through the public API:

1. **`KeyError: 'n'` on standard Client data.** `check_experiment_effects_per_metric` reads `dfm["n"]`, but `n` is not in `Data.REQUIRED_COLUMNS` and is never produced by the standard `Client.complete_trial(...)` flow. Computing the analysis on any such experiment crashes with a bare `KeyError: 'n'` instead of an informative error.

2. **`sem == 0` silently flips the conclusion.** With deterministic data (sem=0, common in BO settings), `no_effect_test_welch` divides by zero (`ws = n / variances`), yielding a NaN p-value. `has_effect = bool(nan < alpha)` is `False`, so the healthcheck warns "no effects have been detected" even when the exactly-known means clearly differ — the opposite of the correct conclusion. The only symptom is a numpy `RuntimeWarning`.

3. **Single-arm trials (K == 1) silently produce NaN.** The Welch statistic divides by `K - 1 == 0`, again giving NaN → `has_effect=False` with no error, even though a one-arm group can't be tested at all.

Also minor: the `check_experiment_effects` docstring describes `ineffective_on_objectives` with the same sentence as `effective` ("can be rejected"), but it means the opposite.

### Please provide a minimal, reproducible example of the unexpected behavior.

```python
from ax import Client, RangeParameterConfig
from ax.analysis.healthcheck.no_effects_analysis import TestOfNoEffectAnalysis

client = Client()
client.configure_experiment(
    parameters=[RangeParameterConfig(name="x", parameter_type="float", bounds=(0, 1))]
)
client.configure_optimization(objective="y")
for x, y in [(0.1, 1.0), (0.5, 2.0), (0.9, 1.5)]:
    idx = client.attach_trial(parameters={"x": x})
    client.complete_trial(trial_index=idx, raw_data={"y": y})

TestOfNoEffectAnalysis().compute(experiment=client._experiment)  # KeyError: 'n'
```

For case 2 (silent wrong result):

```python
import pandas as pd
from ax.core.data import Data
from ax.utils.stats.no_effects import check_experiment_effects_per_metric

data = Data(df=pd.DataFrame([
    {"trial_index": 0, "arm_name": "0_0", "metric_name": "m1",
     "metric_signature": "m1", "mean": 1.0, "sem": 0.0, "n": 100},
    {"trial_index": 0, "arm_name": "0_1", "metric_name": "m1",
     "metric_signature": "m1", "mean": 2.0, "sem": 0.0, "n": 100},
]))
out = check_experiment_effects_per_metric(data=data, objective_names={"m1"})
print(out)  # p_value=NaN, has_effect=False — but means 1.0 vs 2.0 with sem=0
            # is a maximally significant effect
```

### Please paste any relevant traceback/logs produced by the example provided.

```shell
Traceback (most recent call last):
  ...
  File "ax/utils/stats/no_effects.py", line 62, in check_experiment_effects_per_metric
    ns=list(dfm["n"]),
        ...
KeyError: 'n'

# and for sem=0:
ax/utils/stats/no_effects.py:239: RuntimeWarning: divide by zero encountered in divide
ax/utils/stats/no_effects.py:246: RuntimeWarning: invalid value encountered in scalar divide
```

### Ax Version

main (reproduced at d8eeb97c)

### Python Version

3.14

### Operating System

Linux

### (Optional) Describe any potential fixes you've considered to the issue outlined above.

- Validate the `n` column up front in `check_experiment_effects_per_metric` / `check_experiment_effects` and raise a `UserInputError` with an actionable message.
- Handle zero-variance arms explicitly in `no_effect_test_welch`: differing exactly-known means are an exact effect (p=0); identical all-deterministic means are no effect (p=1); a mixed zero/positive-sem group raises `UserInputError` since Welch's test is undefined there.
- Skip single-arm (metric, trial) groups in `check_experiment_effects_per_metric`, validate `K >= 2` and `n > 1` in `no_effect_test_welch`, and raise a `UserInputError` from `TestOfNoEffectAnalysis` when no trial has two or more arms.

I have a patch with these changes plus tests ready and will open a PR.

### Pull Request

Yes, opening one

### Code of Conduct

ricardoofnl:
## Summary

Fixes #5248.

`TestOfNoEffectAnalysis` and the underlying functions in `ax/utils/stats/no_effects.py` mishandled three degenerate inputs:

1. **Data without the optional `n` column** (as produced by standard trial evaluation, e.g. `Client.complete_trial`) crashed with a bare `KeyError: 'n'`. A new `_validate_sample_sizes` helper now raises a `UserInputError` explaining that per-arm sample sizes are required, from both `check_experiment_effects_per_metric` and `check_experiment_effects`.

2. **Deterministic data (`sem == 0`)** made Welch's test divide by zero, producing a NaN p-value that silently read as "no effect" even when the exactly-known means clearly differed — the healthcheck warned "no effects detected" in exactly the wrong situation. Zero-variance arms are now handled explicitly in `no_effect_test_welch`:
   - differing exactly-known means are an exact effect (`p = 0`),
   - identical all-deterministic means are no effect (`p = 1`),
   - a mixed zero/positive-sem group raises a `UserInputError`, since Welch's test is undefined there.

3. **Single-arm groups (`K == 1`)** divided by `K - 1 == 0` and silently produced NaN / "no effect". Such groups are now skipped in `check_experiment_effects_per_metric` (a single-arm group cannot show within-trial effects), `no_effect_test_welch` validates `K >= 2` and `n > 1`, and `TestOfNoEffectAnalysis` raises a `UserInputError` when no trial has two or more arms.

Also corrects the `check_experiment_effects` docstring, which described `ineffective_on_objectives` with the same sentence as `effective`.

## Test plan

- New `ax/utils/stats/tests/test_no_effects.py` covering: missing/null `n` column, `sem == 0` with differing means (exact effect detected), `sem == 0` with equal means (no effect, well-defined p-value), mixed zero/positive sems, single-arm group skipping, and the `K >= 2` / `n > 1` validations.
- Three new cases in `ax/analysis/healthcheck/tests/test_no_effects_analysis.py`: informative error without the `n` column, PASS on deterministic data with real effects (previously a spurious WARNING), and informative error when all trials are single-arm.
- All 18 tests in the two touched test files pass; the end-to-end repro from #5248 now raises an actionable `UserInputError` instead of `KeyError: 'n'`.

### DIFF
--- a/ax/analysis/healthcheck/no_effects_analysis.py
+++ b/ax/analysis/healthcheck/no_effects_analysis.py
@@ -100,6 +100,11 @@ class TestOfNoEffectAnalysis(Analysis):
         df_tone = check_experiment_effects_per_metric(
             data=data, objective_names=set(objective_names)
         )
+        if df_tone.empty:
+            raise UserInputError(
+                "TestOfNoEffectAnalysis requires at least one trial with two "
+                "or more arms, since the test compares arms within a trial."
+            )
         metrics_tone = df_tone.groupby("metric_name")["has_effect"].sum() > 0
         metrics_with_effects = [i for i in metrics_tone.index if metrics_tone[i]]
         objectives_without_effects = set()
--- a/ax/analysis/healthcheck/tests/test_no_effects_analysis.py
+++ b/ax/analysis/healthcheck/tests/test_no_effects_analysis.py
@@ -101,6 +101,74 @@ class TestTestOfNoEffectAnalysis(TestCase):
         with self.assertRaises(AxError):
             self.tone.compute(experiment=self.experiment)
 
+    def test_raises_error_without_sample_sizes(self) -> None:
+        # GIVEN an experiment whose data has no `n` column (as produced by
+        # standard trial evaluation, e.g. `Client.complete_trial`)
+        self.experiment.attach_data(
+            data=Data(
+                df=pd.DataFrame(
+                    {
+                        "arm_name": ["0_0", "0_1", "0_2"],
+                        "metric_name": ["branin"] * 3,
+                        "mean": [0.0, 1.0, 2.0],
+                        "sem": [0.1] * 3,
+                        "trial_index": [0] * 3,
+                        "metric_signature": ["branin"] * 3,
+                    }
+                )
+            )
+        )
+        # WHEN we compute the healthcheck
+        # THEN it raises a UserInputError instead of a KeyError
+        with self.assertRaisesRegex(UserInputError, "per-arm sample sizes"):
+            self.tone.compute(experiment=self.experiment)
+
+    def test_deterministic_data_detects_effects(self) -> None:
+        # GIVEN an experiment with deterministic data (sem == 0) and
+        # clearly different means
+        self.experiment.attach_data(
+            data=Data(
+                df=pd.DataFrame(
+                    {
+                        "arm_name": ["0_0", "0_1", "0_2"],
+                        "metric_name": ["branin"] * 3,
+                        "mean": [0.0, 1.0, 2.0],
+                        "sem": [0.0] * 3,
+                        "trial_index": [0] * 3,
+                        "n": [1000] * 3,
+                        "metric_signature": ["branin"] * 3,
+                    }
+                )
+            )
+        )
+        # WHEN we compute the healthcheck
+        card = self.tone.compute(experiment=self.experiment)
+        # THEN it is a PASS (previously a NaN p-value silently read as
+        # "no effect" and produced a WARNING)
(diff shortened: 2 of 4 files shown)


saitcakmak (approved):
Thanks!

PR #5251, opened 2026-07-09
Fix NaN/Infinity serialization in save_to_json_file (#4144)

### LINKED ISSUE #4144: [Bug]: save_to_json_file saves a value NaN which not compatible with json format
### What happened?

from ax.api.client import Client

When i tried saving the Client to json using save_to_json_file(), it saved an value NaN which is not compatible with json. How to avoid this problem

### Please provide a minimal, reproducible example of the unexpected behavior.

ValueError: Out of range float values are not JSON compliant

"                            "map_key_infos": [
                                {
                                    "key": "step",
                                    "default_value": NaN
                                }
                            ],
"

### Please paste any relevant traceback/logs produced by the example provided.

```shell

```

### Ax Version

1.0.0

### Python Version

3.11.2

### Operating System

debian

### (Optional) Describe any potential fixes you've considered to the issue outlined above.

_No response_

### Pull Request

None

### Code of Conduct

SiddharthSingh9018:
Fixes #4144

## Summary
This makes Ax JSON serialization standards-compliant when non-finite floats appear in the serialized object graph.

- Encodes `nan`, `inf`, and `-inf` using Ax's typed JSON format.
- Handles Python floats and NumPy floating values.
- Recursively handles nested containers, ndarray payloads, tensor payloads, and sets.
- Decodes typed float values back to `nan`, `inf`, and `-inf`.
- Makes `Client.save_to_json_file()` call `json.dumps(..., allow_nan=False)` after conversion.

## Validation
- `python -m ufmt format ax\api\client.py ax\api\tests\test_client.py ax\storage\json_store\encoder.py ax\storage\json_store\decoder.py ax\storage\json_store\decoders.py ax\storage\json_store\tests\test_json_store.py`
- `python -m py_compile ax\api\client.py ax\api\tests\test_client.py ax\storage\json_store\encoder.py ax\storage\json_store\decoder.py ax\storage\json_store\decoders.py ax\storage\json_store\tests\test_json_store.py`
- Direct strict JSON save/load verification passed.

Note: focused pytest collection on Windows is blocked by Ax's `TestCase` using `signal.SIGALRM`, which Windows does not support.

### DIFF
--- a/ax/storage/json_store/decoders.py
+++ b/ax/storage/json_store/decoders.py
@@ -304,6 +304,11 @@ def class_from_json(json: dict[str, Any]) -> type[Any]:
 
 def tensor_from_json(json: dict[str, Any]) -> torch.Tensor:
     try:
+        value = json["value"]
+        if isinstance(value, list):
+            from ax.storage.json_store.decoder import object_from_json
+
+            value = object_from_json(object_json=value)
         device = (
             assert_is_instance(
                 torch_type_from_str(
@@ -315,7 +320,7 @@ def tensor_from_json(json: dict[str, Any]) -> torch.Tensor:
             else torch.device("cpu")
         )
         return torch.tensor(
-            json["value"],
+            value,
             dtype=assert_is_instance(
                 torch_type_from_str(
                     identifier=json["dtype"]["value"], type_name="dtype"
--- a/ax/storage/json_store/decoders.py
+++ b/ax/storage/json_store/decoders.py
@@ -305,10 +305,6 @@ def class_from_json(json: dict[str, Any]) -> type[Any]:
 def tensor_from_json(json: dict[str, Any]) -> torch.Tensor:
     try:
         value = json["value"]
-        if isinstance(value, list):
-            from ax.storage.json_store.decoder import object_from_json
-
-            value = object_from_json(object_json=value)
         device = (
             assert_is_instance(
                 torch_type_from_str(
--- a/ax/storage/json_store/encoder.py
+++ b/ax/storage/json_store/encoder.py
@@ -9,6 +9,7 @@
 import dataclasses
 import datetime
 import enum
+import math
 from collections import OrderedDict
 from collections.abc import Callable
 from functools import partial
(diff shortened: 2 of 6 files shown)


ax/storage/json_store/decoders.py
```diff
@@ -304,6 +304,11 @@ def class_from_json(json: dict[str, Any]) -> type[Any]:
 
 def tensor_from_json(json: dict[str, Any]) -> torch.Tensor:
     try:
+        value = json["value"]
+        if isinstance(value, list):
+            from ax.storage.json_store.decoder import object_from_json
```
saitcakmak: I'd like to avoid this circular import. For other cases like numpy arrays, we often decode the underlying object in `object_from_json` before calling the decoder from another file (like `tensor_from_json`). Let's repeat this pattern and instead call `object_json["value"] = _object_from_json(object_json["value"])` before passing it down here. That will also generalize this handling to 0-dim tensors. Those are currently not handled here due to `isinstance(value, list)` check.

ax/storage/json_store/encoder.py
```diff
@@ -115,11 +117,11 @@ def object_to_json(
...
         return {"__type": _type.__name__, "name": obj.name}
     elif _type is np.ndarray or issubclass(_type, np.ndarray):
-        return {"__type": _type.__name__, "value": obj.tolist()}
+        return {"__type": _type.__name__, "value": _object_to_json(obj.tolist())}
     elif _type is set:
-        return {"__type": _type.__name__, "value": list(obj)}
+        return {"__type": _type.__name__, "value": _object_to_json(list(obj))}
     elif _type is torch.Tensor:
-        return tensor_to_dict(obj=obj)
+        return {k: _object_to_json(v) for k, v in tensor_to_dict(obj=obj).items()}
```
saitcakmak: This is introducing a python loop over each element of arrays / tensors. It could be worth gating the `_object_to_json` calls behind checks for inf / nan elements to avoid paying the cost for all objects

saitcakmak (commented):
With this, we're intro

ax/storage/json_store/encoder.py
SiddharthSingh9018: Thanks,
 For the **circular import**, I'll follow the existing pattern and move the recursive decode of object_json["value"] into object_from_json in decoder.py, before dispatching to decoders.py. 
That also fixes the 0-dim tensor case you flagged, since the recursion won't depend on isinstance(value, list) anymore.

Plan:
Update decoder.py / decoders.py per the circular import fix, remove the local import
Add a test for a 0-dim tensor containing NaN
For the encoder perf concern: add a check that scans the array first to see if it actually contains any NaN/Inf

If all values are normal numbers (common case) → skip per-element conversion, use the existing fast path
If NaN/Inf is present (rare case) → fall back to per-element conversion
Skip the check entirely for non-float dtypes, since NaN/Inf doesn't apply there
Leave sets on a simple loop, since there's no vectorized check available

SiddharthSingh9018:
Updated in 67948da, addressing both review comments:
- moved recursive value decoding into object_from_json to remove the tensor decoder circular import and cover 0-dimensional tensor/array values
- kept finite ndarray/tensor/set serialization on the fast path while only falling back to per-element conversion for non-finite values
I also added regression coverage for 0-dimensional non-finite ndarray/tensor values and finite ndarray/tensor output stability.

saitcakmak (approved):
Thanks!

PR #5289, opened 2026-09-12
Decode graphviz SVG bytes in analysis card HTML body

rootkiller6788:
Graphviz cards (the Generation Strategy Graph you get from compute_analyses, and the hierarchical search space one) were showing a wall of raw text instead of a graph in notebooks.

graphviz's Source.pipe returns bytes unless you pass an encoding, and the card was embedding that bytes object straight into its HTML div - so the browser got something like b'<?xml ...' as text rather than an <svg> element to render.

Decode the pipe output to a string before putting it in the HTML. Also added a small regression test that mocks pipe to return bytes and checks the body actually contains an <svg> tag.


saitcakmak (approved):
thanks for the fix!

